PR #8, opened 2023-03-20
Feat/unstoppable domains

t0rbik:
Add unstoppable domains resolution.


t0rbik:
Await falks to confirm it works for them and maybe new API key

PR #21, opened 2023-04-02
Feature/show total bribes

oxSaturn:
Hey, I think it might be a good idea to show the bribes accrued every week.

Let me know what you think.

### DIFF
--- a/Frontend-v1-Original/components/ssVotes/ssVotes.tsx
+++ b/Frontend-v1-Original/components/ssVotes/ssVotes.tsx
@@ -249,7 +249,12 @@ export default function ssVotes() {
     );
   };
 
-  const filteredGauges = useMemo(
+  interface Row {
+    gauge: {
+      tbv: number
+    }
+  }
+  const filteredGauges: Row[] = useMemo(
     () =>
       gauges.filter((pair) => {
         if (!search || search === "") {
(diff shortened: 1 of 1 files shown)


Frontend-v1-Original/components/ssVotes/ssVotes.tsx
```diff
@@ -249,7 +249,12 @@ export default function ssVotes() {
     );
   };
 
-  const filteredGauges = useMemo(
+  interface Row {
+    gauge: {
+      tbv: number
+    }
+  }
+  const filteredGauges: Row[] = useMemo(
```
t0rbik, line 257: This type casting is wrong. In production branch filteredGauges inherits Gauge[] type which has gauge.tbv in it as a required field. Moreover, typescript wouldn't allow you to pass this filtered gauges to GaugesTable, because it awaits type Gauge[]:
```
<GaugesTable
          gauges={filteredGauges}
          setParentSliderValues={setVotes}
          defaultVotes={votes}
          token={token}
/>
```

```suggestion
  const filteredGauges = useMemo(
```

Frontend-v1-Original/components/ssVotes/ssVotes.tsx
```diff
@@ -276,21 +276,36 @@ export default function ssVotes() {
...
+        acc: number,
+        row: {
+          gauge: {
+            tbv: number;
+          };
+        }
+      ) => acc + row.gauge.tbv,
+      0
+    )
+    .toFixed(2);
```
oxSaturn, line 291:
```suggestion
    .toFixed(2)
    .toLocaleString();
```
Might be a good idea to format large numbers with `toLocaleString`?

E.g., showing `$20,032.8` instead of `$20032.80`.

t0rbik: @oxSaturn sure. to be consistent you can import function called formatCurrency
`import { formatCurrency } from "../../utils/utils";`
 and do something like 
`...{$formatCurrency(bribesAccrued)}...`

t0rbik:
merged @oxSaturn

PR #24, opened 2023-04-04
Introduce viem public client for blockchain reading

t0rbik:
This is a WIP.
This PR will introduce viem.sh public client for fetching blockchain using fallback provider (it will consist of all known canto rpc urls) to not rely on users rpc.


t0rbik:
The fallback Transport consumes multiple Transports. If a Transport request fails, it will fall back to the next one in the list.

t0rbik:
Reference on using contract instances:

> While Contract Instances are great for reducing code duplication, they pull in multiple contract actions (e.g. createContractEventFilter, estimateContractGas, readContract, simulateContract, watchContractEvent, writeContract), so they can be a bit heavier than individual calls. If you only need a couple contract methods and you care about minimizing bundle size to the fullest extent, you may want to use individual calls instead.

[docs](https://viem.sh/docs/contract/getContract.html)

PR #25, opened 2023-04-05
add api for getting flow circulating supply on canto

oxSaturn:
So this pull request will add a new api at `https://www.velocimeter.xyz/api/canto/circulating-supply` which would return data per coingecko's requirements https://docs.google.com/document/d/1v27QFoQq1SKT3Priq3aqPgB70Xd_PnDzbOCiuoCyixw/edit#heading=h.pydh6ytf8qyc

I'm using viem here because I noticed that @t0rbik is refactoring code with viem in https://github.com/Velocimeter/frontend/tree/breaking/viem.


Frontend-v1-Original/pages/api/canto/circulating-supply.ts
```diff
@@ -0,0 +1,312 @@
...
+const publicClient = createPublicClient({
+  chain: canto,
+  transport: http(),
+});
+
+const flowContract = getContract({
+  address: CONTRACTS.GOV_TOKEN_ADDRESS as `0x${string}`,
+  // I'm inlining the abi here to enable type inference for typescript
+  // see https://viem.sh/docs/typescript.html#type-inference
+  abi: [
```
oxSaturn, line 15: Note that I'm not importing the abi here because the abi is in a `.js` file instead of `.ts`. And enabling type inference would require rewriting `.js` to `.ts` and append a `as const` which would cause a lot of typing errors at the moment.

Frontend-v1-Original/pnpm-lock.yaml
t0rbik, line 1: Can we remove it? we dont use pnpm, we use yarn

Frontend-v1-Original/pages/api/canto/circulating-supply.ts
```diff
@@ -0,0 +1,332 @@
+import type { NextApiRequest, NextApiResponse } from "next";
+import { createPublicClient, getContract, http } from "viem";
+import { canto } from "viem/chains";
+import { CONTRACTS, NATIVE_TOKEN } from "../../../stores/constants/constants";
+import Cors from "cors";
+const cors = Cors({
+  methods: ["GET"],
+});
+const publicClient = createPublicClient({
```
t0rbik, line 9: Could you initialize publicClient same way as here [this](https://github.com/Velocimeter/frontend/blob/breaking/viem/Frontend-v1-Original/stores/connectors/viem.ts)?

t0rbik: So we can rely on multiple rpc urls

Frontend-v1-Original/pages/api/canto/circulating-supply.ts
t0rbik: sounds good. we will make it better after #24 gets merged

Frontend-v1-Original/pages/api/canto/circulating-supply.ts
```diff
@@ -0,0 +1,332 @@
...
+  // so we're safe to use Number to convert Bigint here
+  const circulatingSupply =
+    Number(
+      totalSupply -
+        lockedSupply -
+        flowInMinter -
+        flowInMsig -
+        flowInRewardsDistributor -
+        flowInTimelockerController
+    ) / Math.pow(10, NATIVE_TOKEN.decimals);
```
t0rbik, line 329: can we use [this ](https://github.com/Velocimeter/frontend/blob/0e96d5606d431efb205b38fa1c775f9f44043eb1/Frontend-v1-Original/stores/helperStore.ts#L81) code for consistency?

t0rbik (changes requested):

Frontend-v1-Original/pages/api/canto/circulating-supply.ts
oxSaturn: If we're not urgent to get this api live, how about continue it after #24 so we don't have to duplicate code?

t0rbik: > If we're not urgent to get this api live, how about continue it after #24 so we don't have to duplicate code?

1. its kind of urgent because CG guys are waiting for it.
2. not sure #24 will be ready to merge today.
3. next apu is kind of isolated so i say we dont worry about duplicated code for now

gonna autorise vercel, lets see if it works the way it is rn

oxSaturn: Updated.

t0rbik:
@oxSaturn hey im thinking rn...
can we do this endpoint to be accessible only from exactly defined origins? like we're going to expose it to public and it creates possibility of abstract attacker doing ddos or something related

oxSaturn:
> @oxSaturn hey im thinking rn... can we do this endpoint to be accessible only from exactly defined origins? like we're going to expose it to public and it creates possibility of abstract attacker doing ddos or something related
> 
> [Array - set origin to an array of valid origins. Each origin can be a String or a RegExp. For example ["http://example1.com", /.example2.com$/] will accept any request from "http://example1.com/" or from a subdomain of "example2.com".](https://github.com/expressjs/cors#configuration-options)

yeah, we can definitely set the `origin`. Do you know which origin Coingecko uses to fetch the data?

t0rbik:
Waiting GC to send over their origin

Frontend-v1-Original/pages/api/canto/circulating-supply.ts
```diff
@@ -1,14 +1,25 @@
...
+const nodestake = http("https://jsonrpc.canto.nodestake.top");
+const slingshot = http("https://canto.slingshot.finance");
+const neobase = http("https://canto.neobase.one");
+
 const publicClient = createPublicClient({
   chain: canto,
-  transport: http(),
+  transport: fallback([dexvaults, plexnode, nodestake, slingshot, neobase], {
+    rank: {
+      interval: 12_000,
```
t0rbik, line 21: Assuming themselved are on pro CG api lets set it to 30_000
Here's reference in CG api:
This depends - but endpoints are all cached to around 1 to 5 minutes and you can expect most data to be updated at similar intervals unless otherwise specified. Pro API (paid plans) generally have faster update frequency, i.e. 30 sec for simple/price endpoint

t0rbik: @oxSaturn i see now they are polling every 30 mins. so we can make it 900_000

Frontend-v1-Original/pages/api/canto/circulating-supply.ts
```diff
@@ -0,0 +1,312 @@
...
+  const circulatingSupply =
+    Number(
+      totalSupply -
+        lockedSupply -
+        flowInMinter -
+        flowInMsig -
+        flowInRewardsDistributor -
+        flowInTimelockerController
+    ) / Math.pow(10, NATIVE_TOKEN.decimals);
+
```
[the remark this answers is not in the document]
t0rbik: @oxSaturn dont need it if they polling every 30 mins

t0rbik:
@oxSaturn let me know if u will make any changes or i will finalise it myself

Frontend-v1-Original/pages/api/canto/circulating-supply.ts
```diff
@@ -5,6 +5,7 @@ import { CONTRACTS, NATIVE_TOKEN } from "../../../stores/constants/constants";
 import Cors from "cors";
```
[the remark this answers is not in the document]
oxSaturn: Better to enforce the styling with an eslint plugin when possible, e.g., https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/newline-after-import.md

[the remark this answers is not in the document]
t0rbik: obv on the list. as well as `run next lint`.

t0rbik (approved):
looks good

t0rbik (approved):

t0rbik:
merged

PR #30, opened 2023-04-07
tvl in liq pairs table

t0rbik:


oxSaturn:
How about appending tvl to `Total Pool Amount` column so we don't have to introduce another column which would be a headache to responsiveness? Something like this:

t0rbik:
> How about appending tvl to `Total Pool Amount` column so we don't have to introduce another column which would be a headache to responsiveness? Something like this:
> 
> 

Yes! I like it. Maybe we can make it +100 bolder to distinguish?

oxSaturn:
> > How about appending tvl to `Total Pool Amount` column so we don't have to introduce another column which would be a headache to responsiveness? Something like this:
> > 
> 
> Yes! I like it. Maybe we can make it +100 bolder to distinguish?

haha, I don't have any strong idea about that, feel free to use what you think is suitable. Or we can just move tvl to top as users might care about it more than other two data.

PR #32, opened 2023-04-08
optimize UI styles

oxSaturn:
Hi, I'd like to make some changes to the UI, let me know what you think.


t0rbik:
@oxSaturn I like offset changes, but at the same time I like cantoGreen colout better

Frontend-v1-Original/components/ssLiquidityManage/ssLiquidityManage.tsx
```diff
@@ -1332,7 +1332,7 @@ export default function ssLiquidityManage() {
     <div className="relative">
       <Paper
         elevation={0}
-        className="m-auto mt-0 flex w-[calc(100%-60px)] max-w-[485px] flex-col items-start bg-transparent shadow-lg shadow-cantoGreen"
+        className="m-auto mt-0 flex w-[calc(100%-60px)] max-w-[485px] flex-col items-start bg-transparent shadow-glow shadow-cantoGreen"
```
t0rbik, line 1335: you don't need shadow-cantoGreen, because shadow-glow already has it in tailwind config. We can rename shadow-glow to shadow-glowCantoGreen in this case so we know it's glowing with cantoGreen color.

Frontend-v1-Original/components/ssSwap/ssSwap.tsx
```diff
@@ -6,7 +6,7 @@ function Swap() {
     <div className="mt-32 flex h-full min-h-[calc(100vh-432px)] w-full flex-col items-center justify-evenly sm:mt-0 lg:flex-row">
       <Paper
         elevation={0}
-        className="flex w-full max-w-[485px] flex-col bg-transparent p-3 shadow-lg shadow-cantoGreen lg:p-6"
+        className="flex w-full max-w-[485px] flex-col bg-transparent p-3 shadow-glow shadow-cantoGreen lg:p-6"
```
t0rbik, line 9: you don't need shadow-cantoGreen, because shadow-glow already has it in tailwind config. We can rename shadow-glow to shadow-glowCantoGreen in this case so we know it's glowing with cantoGreen color.

```suggestion
        className="flex w-full max-w-[485px] flex-col bg-transparent p-3 shadow-glowCantoGreen lg:p-6"
```

Frontend-v1-Original/components/ssLiquidityManage/ssLiquidityManage.tsx
t0rbik:
```suggestion
        className="m-auto mt-0 flex w-[calc(100%-60px)] max-w-[485px] flex-col items-start bg-transparent shadow-glowCantoGreen"
```

Frontend-v1-Original/tailwind.config.js
```diff
@@ -45,6 +45,9 @@ module.exports = {
       colors: {
         cantoGreen: "#06fc99",
       },
+      boxShadow: {
+        glow: "0 0 10px 0 #06fc99",
```
t0rbik, line 49:
```suggestion
        glowCantoGreen: "0 0 10px 0 #06fc99",
```

t0rbik (approved):
Hey. I left a suggestion. On the other hand I feel that your way is also ok because it's following default tailwind `shadow-_size_ shadow-_color_` way of doing box-shadow

t0rbik:
wyt @oxSaturn

oxSaturn:
> Hey. I left a suggestion. On the other hand I feel that your way is also ok because it's following default tailwind `shadow-_size_ shadow-_color_` way of doing box-shadow

I would suggest keeping consistence with the naming convention of tailwind.

All `shadow-` classes in tailwindcss has a default color:



In my code, we happened to have `cantoGreen` as the default color.

oxSaturn:
If we change it to `shadow-glowCantoGreen`, we'll have to define another className when we need another color, right? We wouldn't be able to compose it with other `shadow-color`.

t0rbik (approved):

t0rbik:
merged

PR #35, opened 2023-04-10
Feature/non web3 browsers

oxSaturn:
Closes https://app.dework.xyz/velocimeter/main-space-58499?taskId=2bd2a946-168a-4c56-8493-d465edc0bfe5

Here's what I did in this pull request:

1. Cover only the bottom half of the screen so users should be able to interact with the top half of the site even though they're not connected

    

3. `dispatch` `ACTIONS.CONFIGURE_SS` in `_app` so data for `Info` could be loaded without switching to canto network first which can't be done on non web3 browsers, i.e., we can see data below now on non web3 browsers:
    
    

5.  disable `useOnClickOutside` as clicking the menu button actually has two side effects: 1. click outside to hide the dropdown, 2. show the dropdown, those two are mutually conflicting IMO.


Frontend-v1-Original/pages/_app.tsx
```diff
@@ -51,6 +51,9 @@ export default function MyApp({
     stores.emitter.on(ACTIONS.ACCOUNT_CONFIGURED, accountConfigureReturned);
 
     stores.dispatcher.dispatch({ type: ACTIONS.CONFIGURE });
+    stores.dispatcher.dispatch({
+      type: ACTIONS.CONFIGURE_SS,
+    });
```
t0rbik, line 56: i will double check this with what we have in account store.

t0rbik: tbh, i want to refactor all of the flux-stores to use react-query (fetching pairs etc with easy recaching), web3react to wagmi+rainbowkit and probably some state in zustand. but yeh, not a 1 day ticket.
Also flux is the last thing keeping us from updating react v to 18 and next v to 13.

oxSaturn: great to hear there's a plan. flux is too old and not maintained anymore as far as I know.

t0rbik: @oxSaturn do you think we could refactor keeping
[this](https://github.com/Velocimeter/frontend/blob/88d8be4f36d8fdf5b11baa626df44129806c687f/Frontend-v1-Original/stores/accountStore.ts#L128 )
in mind?
maybe move this to else statement of that `if (isAuthorized && isChainSupported)`

oxSaturn: Sounds good.

Frontend-v1-Original/components/header/mobileHeader.tsx
```diff
@@ -299,14 +299,16 @@ function Header() {
...
+              open ? <CloseIcon /> : <MenuIcon />
+            }
           </button>
         </div>
         <div
           ref={ref}
           className={`${
             open ? "" : "hidden"
-          } absolute top-[105%] right-2 z-10 flex w-80 animate-slideLeftAndFade flex-col items-start justify-between rounded-sm border-none bg-[#0d3531] py-5 px-6 shadow-md shadow-cantoGreen`}
+          } absolute top-[105%] right-2 z-10 flex w-80 animate-slideLeftAndFade flex-col items-start justify-between rounded-sm border-none bg-[#0d3531] py-5 px-6 shadow-glow shadow-cantoGreen`}
```
t0rbik, line 311: we can delete shadow-cantoGreen

oxSaturn:
Reckon the styles of overlay can be improved further :) but I have no good idea at the moment.

t0rbik:
@oxSaturn we should probably worry about style of this overlay too much hehe. 
I mean I want to merge and close it already, but I'm thinking rn. Are we sure, that one wouldn't be able to sign any transaction on canto using account not connected to canto?

For instance, with this overlay, I can go to Vest page, click on my first nft, reset it and injected wallet will try to send transactions using my connected chain in wallet (see attached screenshot).

I understand that possibility of them using app while seeing big overlay literally saying 'Switch chain' is not high, however, we should think of it I reckon. 

Any easy way we could implement this overlay to be actually not overlay but in-page? Like it would replace page data same way as Connect Wallet replaces it right now.

image

t0rbik:
@oxSaturn 
image
ref1

image
ref2

t0rbik:
So another solution is leave this Switch Network as it is but bring wallet connect modal opener button there

oxSaturn:
I see the problem you described.

Here's what I think, since there're two user actions needed:

1. switch chain
2. connect wallet

Why not just combining them into one click?

I've made some changes to the `/swap` page so clicking the `Connect Wallet to Continue` button will do both. And I've tested the function under both Metamask and Rabby.

Let me know if it makes sense so I can continue to update other buttons.

oxSaturn:
Oops, seems it won't solve the problem you described since I refresh the page every time I change the network. Will need further fix, let me convert it into draft first, will ping you once it's ready.

t0rbik:
Now if I connected to dapp, switched to canto chain the Unlock goes away. But if I switch to Optimism nothing pops up, meaning I'm able to sign txs again (screenshot attached).
image

t0rbik:
> Oops, seems it won't solve the problem you described since I refresh the page every time I change the network. Will need further fix, let me convert it into draft first, will ping you once it's ready.

Ok, would be awesome to fix it. As I see our colleagues did come up with their solutions (I mean refs I've sent earlier).

oxSaturn:
> Another thing is I can't connect wallet if I'm actually connected, but on another chain

This seems to be a bug reproducible on production, just being exposed due to the changes introduced in this PR.

Frontend-v1-Original/components/common/OneClickButton.tsx
```diff
@@ -0,0 +1,83 @@
+import { Button, Typography } from "@mui/material";
+import { useChainInvalid } from "../../hooks/useChainInvalid";
+import UnlockModal from "../unlock/unlockModal";
+import { useEffect, useState } from "react";
+import { switchChain } from "../header/header";
+import stores from "../../stores";
+import { ACTIONS } from "../../stores/constants/constants";
+import { useAccount } from "../../hooks/useAccount";
+import { EthWindow } from "../../stores/types/types";
```
oxSaturn, line 9: Too many duplicated code here and there, so I refactored the shared part into this `OneClickButton` component.

Frontend-v1-Original/components/common/OneClickButton.tsx
```diff
@@ -0,0 +1,83 @@
...
+    setUnlockOpen(false);
+  };
+  const onAddressClicked = async () => {
+    if (chainInvalid) {
+      const chainSwitched = await switchChain();
+      // user might cancel the switch
+      if (chainSwitched !== false && !(account && account.address)) {
+        if ((window as EthWindow).ethereum?.isRabby === true) {
+          // rabby will connect and switch network in one step
+          // hence we don't need to open the unlock modal
```
oxSaturn, line 36: It's a hack here to eliminate the flash of unlock modal as rabby would connect and switch network automatically in one step instead of two in MetaMask.

Frontend-v1-Original/components/header/header.tsx
```diff
@@ -339,27 +343,6 @@ function Header() {
...
-              onClick={() => switchChain()}
-            >
-              Switch to{" "}
-              {process.env.NEXT_PUBLIC_CHAINID == "740"
-                ? "Canto testnet"
-                : "Canto"}
-            </Button>
-          </div>
-        </div>
-      ) : null}
```
oxSaturn, line 362: The overlay is totally removed now.

oxSaturn:
@t0rbik so this pull request made quite a lot of changes now, the most significant one is the overlay was removed.

I found it distracting, and I believe users shouldn't be worry about the wrong network, we can make sure they will jump to the right network when they're connecting to the site.

Just too many duplicated code everywhere and I can't help refactoring more and more of them, lol, anyway, please review and let me know what you think.

Frontend-v1-Original/stores/accountStore.ts
```diff
@@ -91,12 +91,18 @@ class Store {
 
   configure = async () => {
     // this.getGasPrices();
-    injected.isAuthorized().then((isAuthorized) => {
+    injected.isAuthorized().then(async (isAuthorized) => {
```
t0rbik, line 94: why?

Frontend-v1-Original/stores/accountStore.ts
```diff
@@ -91,12 +91,18 @@ class Store {
...
-      // fall back to canto mainnet if chainId undefined
-      const { chainId = process.env.NEXT_PUBLIC_CHAINID } =
-        (window as EthWindow).ethereum || {};
+      // see https://github.com/solidex-fantom/solidly-ui/pull/3/files
+      let chainId;
+      if ((window as EthWindow).ethereum) {
+        chainId = await (window as EthWindow).ethereum.request({
+          method: "eth_chainId",
+        });
+      } else {
```
t0rbik, line 103: got it

t0rbik (approved):
let me deploy and test manually a bit. seems good

t0rbik:
about the repeating code... Seems like original code is literally copy pasting pages/components.
Regarding account store imperativism, I reckon wagmi devs are going to release wagmi with viem underhood in the next fortnight and we can grab it as soon as it's stable (because we already have viem).

t0rbik:
for sanity check, does it help with non web3 browsers in the end?

t0rbik:
seems like I can't connect with wallet connect on non-web3 browser

oxSaturn:
> for sanity check, does it help with non web3 browsers in the end? seems like I can't connect with wallet connect on non-web3 browser

I can confirm it doesn't work on velocimeter.xyz as well. I can use Velodrome on non web3 browsers with WalletConnect , seems we have more to fix before landing this pr.

oxSaturn:
So maybe we can then upgrade the web3 react connector to the ones recommended here https://wagmi.sh/examples/connect-wallet ? We can put this pr on hold until then.

t0rbik:
> > for sanity check, does it help with non web3 browsers in the end? seems like I can't connect with wallet connect on non-web3 browser
> 
> I can confirm it doesn't work on velocimeter.xyz as well. I can use Velodrome on non web3 browsers with WalletConnect , seems we have more to fix before landing this pr.

yeah one of the reasons is that velodrome uses wagmi, judging by rainbow kit wallet connect button. Im guessing but i dont think they still use this imperative code.



> So maybe we can then upgrade the web3 react connector to the ones recommended here https://wagmi.sh/examples/connect-wallet ? We can put this pr on hold until then.

for sure we can wait a bit

t0rbik:
dropping viem+wagmi PR here [PR2135](https://github.com/wagmi-dev/wagmi/pull/2135)

t0rbik:
Closing in favour of #70

PR #36, opened 2023-04-10
Various UI Fixes + Improvements

shatterproof:
1. Adjusted the meta description to use Canto instead of Arbitrum
2. Changed references from MIM token to NOTE (MIM does not exist on Canto)
3. Adjusted format of the token list JSON attributes to be in a consistent order
4. Added proper names for tokens to help with user searching
5. Updated icons for multiBTC, cINU, and CRE8R tokens
6. Fixed small token images not using the full width on the selection screen (OSMO + SOMM tokens appeared smaller than the others)
7. Fixed a minor typo on the home page (NFT's → NFTs)

Preview of some of the changes (old vs. new):
Old → New

edit: reverted multiBTC icon back to the original to match the bridge


Frontend-v1-Original/mainnet-canto-token-list.json
t0rbik, line 1: @shatterproof just double cheking: have you checked all of these tokens names?
Another thing is we shouldn't change multiBTC logo at least before multichain folks change it on their frontend, because you first need to go through their frontend to bridge to canto.

t0rbik (changes requested):

t0rbik (approved):

t0rbik (dismissed):

t0rbik (approved):

PR #37, opened 2023-04-10
hotfix bribes not showing

t0rbik:


t0rbik:
@oxSaturn hey, can you think of any good way adopting multicall on chunks in here? i ended up just hotfixing it, because bribes werent showing properly. 
Promise.all works for now, but I can expect it to become not efficient enough when we have more pairs. 1 Pair has at least 2 bribes to check -> lots of Promises.
I will think of it as well. Sharing thoughts.

t0rbik:
#39 fixes it properly

PR #38, opened 2023-04-11
Feat/vote manager

t0rbik:


t0rbik:
This PR will introduce Vote Manager feature.
User will be able to delegate and undelegate their veNFT to Flow Vote Manager Contract. User will be able to enable or disable an autolock feature, which is to extend your veFLOW lock duration by 1 week every week to maintain voting power.

This PR will also introduce Zustand store as well as React-Query. It will set an example of how to use Zustand stores playing alongside React-Query. This example will be used to gradually migrate from Flux, which will allow us to update React and Next.js to latests versions.

Issue #75, opened 2023-05-03
Wrong Reward Estimate for voted nft


oxSaturn:
Say I have a voted nft, now if I drag the slider to estimate reward of voting other pairs, the `Reward Estimate` could return wrong value:

![](https://user-images.githubusercontent.com/126733611/235842716-96f26496-a715-470a-a973-52336e9300a3.png)

As you can see from the screenshot, the total bribes value is only `$0.04` while `Reard Estimate` returns `$979.24`.

t0rbik:
Merged #81. Also it explicitly says estimate.
According to [Cambridge dictionary](https://dictionary.cambridge.org/dictionary/english/estimate) estimate is:

> a **guess** of what the size, value, amount, cost, etc. of something might be.

> a **judgment** or calculation of **approximately** how large or how great something is.

t0rbik:
Although in regard to the attached screenshot it is fair to say that the situation is more of a rough estimate.

PR #81, opened 2023-05-05
grouping nfts

oxSaturn:
Currently there's no way to tell right from the vote page which nfts have voted or reset:

You'll have to check the vest page first.

In this pr, we group nfts into two categories:

And if you select those voted/reset nfts:

We'll show this warning:

Instead of an actionable button like below:

And since users can tell right on the vote page the status of nfts, they're unlikely to do something described in https://github.com/Velocimeter/frontend/issues/75, hence we can hopefully avoid that issue.

---

## PR-Codex overview
This PR adds a `MyListSubheader` component and an `Alert` component, and implements logic to filter and display only votable NFTs. It also adds an error message when a user tries to vote or reset an NFT that has already been actioned in the current epoch.

### Detailed summary
- Adds `MyListSubheader` component
- Adds `Alert` component
- Filters and displays only votable NFTs
- Adds error message when attempting to vote/reset an NFT that has already been actioned in the current epoch


Frontend-v1-Original/components/ssVotes/ssVotes.tsx
```diff
@@ -234,8 +256,49 @@ export default function Votes() {
...
+                  {votableNFTs.length > 0 ? (
+                    <MyListSubheader>
+                      Available to Vote this Epoch
+                      <ExpandMore />
+                    </MyListSubheader>
+                  ) : null}
+                  {votableNFTs.length > 0 &&
+                    votableNFTs.map((option) => {
+                      return (
+                        <MenuItem
```
oxSaturn, line 268: That's why I'm not introducing a customized component to avoid code duplication.

Frontend-v1-Original/components/ssVotes/ssVotes.tsx
```diff
@@ -41,6 +44,12 @@ const initialEmptyToken: VestNFT = {
   lastVoted: BigInt(0),
 };
 
+function MyListSubheader(props: ListSubheaderProps) {
+  return <ListSubheader {...props} className="" />;
+}
+
+MyListSubheader.muiSkipListHighlight = true;
+
```
oxSaturn, line 52: See https://mui.com/material-ui/react-select/#grouping

Frontend-v1-Original/components/ssVotes/ssVotes.tsx
```diff
@@ -76,7 +85,14 @@ export default function Votes() {
...
     if (nfts && nfts.length > 0) {
-      setToken(nfts[0]);
+      const votableNFTs = nfts.filter(
+        (nft) => nft.actionedInCurrentEpoch === false
+      );
+      if (votableNFTs.length > 0) {
+        setToken(votableNFTs[0]);
+      } else {
+        setToken(nfts[0]);
+      }
```
oxSaturn, line 95: Select votable NFTs first when possible.

t0rbik (approved):
looks very good

t0rbik:
When i first open vote page with 1 nft voted and 1 nft hasnt voted. it auto selects nft not voted in nft selector. But the ui shows votes for nft voted. because it has lower id so it comes first in array I guess and in prod it would select first my voted nft.

t0rbik (changes requested):
When i first open vote page with 1 nft voted and 1 nft hasnt voted. it auto selects nft not voted in nft selector. But the ui shows votes for nft voted. because it has lower id so it comes first in array I guess and in prod it would select first my voted nft.

t0rbik (approved):
Awesome. Do you reckon we can delete expand more icon because it's not expandable?

oxSaturn:
> Awesome. Do you reckon we can delete expand more icon because it's not expandable?

The reason I added the expand more icon was to distinguish the heading since the default style sucks. But yeah, it would introduce ambiguity here.

PR #100, opened 2023-05-25
show influence column for vest page

oxSaturn:
The purpose the influence is to help track my share of the total votes. Currently I had to do the math manually myself.

Since there's another column introduced, I would like to combine all the action buttons into one menu so they don't take too much space. And we might include more actions in future like `transfer`ing nfts.

---

## PR-Codex overview
This PR adds a new `influence` field to the vesting NFTs and displays it in the vesting table. It also includes various code changes related to this feature.

### Detailed summary
- Added `influence` field to vesting NFTs
- Display `influence` in vesting table
- Various code changes related to `influence` feature

> The following files were skipped due to too many changes: `Frontend-v1-Original/components/ssVests/ssVestsTable.tsx`

### DIFF
--- a/Frontend-v1-Original/components/ssRewards/ssRewards.tsx
+++ b/Frontend-v1-Original/components/ssRewards/ssRewards.tsx
@@ -30,6 +30,7 @@ const initialEmptyToken: VestNFT = {
   actionedInCurrentEpoch: false,
   reset: false,
   lastVoted: BigInt(0),
+  influence: 0,
 };
 
 export default function Rewards() {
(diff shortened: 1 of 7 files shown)


t0rbik:
Can we pick something else for extend icon?

oxSaturn:
> Can we pick something else for extend icon?

Any suggest? What about this one:

t0rbik:
What you think? just went through all lol. yours also works. Also is "Extend" clear that it's increase and extend time both. Wouldn't  better to leave it as "Manage"?
Screenshot 2023-05-28 at 6 00 39 pm
Screenshot 2023-05-28 at 6 03 46 pm

oxSaturn:
> What you think? just went through all lol. yours also works. Also is "Extend" clear that it's increase and extend time both. Wouldn't better to leave it as "Manage"? Screenshot 2023-05-28 at 6 00 39 pm Screenshot 2023-05-28 at 6 03 46 pm

In the best case, we could separate Increase lock amount and Increase duration into two menu items. But maybe better keeping `Manage` at the moment.

t0rbik (approved):

PR #107, opened 2023-05-30
add few styles to toggles and delete gauge filter

t0rbik:
## PR-Codex overview
This PR adds a new filter feature to the `ssLiquidityPairsTable` component and refactors the toolbar. 

### Detailed summary
- Added a new filter feature to the `ssLiquidityPairsTable` component
- Refactored the toolbar to include the new filter
- Removed toggle buttons for filtering by My Deposits, Active Gauge, Stable, Volatile, and Boosted
- Changed local storage keys from `pairsToggle-v1` to `pairsState-v1`

### DIFF
--- a/Frontend-v1-Original/components/ssLiquidityPairs/ssLiquidityPairsTable.tsx
+++ b/Frontend-v1-Original/components/ssLiquidityPairs/ssLiquidityPairsTable.tsx
@@ -17,7 +17,6 @@ import {
   IconButton,
   TextField,
   InputAdornment,
-  Switch,
   Collapse,
   List,
   ListItem,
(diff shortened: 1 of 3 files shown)


oxSaturn:
Looks good. But I'm wondering if we can use radio group like below:



You select only one category each time (Will user really needs to select boosted stable?).

Currently, there're actually multiple composition:

1. Stable ∪ Volatile
2. Deposited ∩ (Stable ∪ Volatile ∪ Boosted)
3. etc.

Which I found confusing sometimes.

oxSaturn:
btw, I believe this is not vertically aligned:

oxSaturn:
Noticed that we don't show these pools before.

oxSaturn:
Another issue I found is, say I'm viewing volatile pools, page 3, now I click stable pools, but there's no page 3 for stable pools, and I can only see empty table. My question is, shall we reset the page to 1 in this case?

t0rbik:
Pairs without gauges were always there, they just filtered out by default. I don't see any reason for filtering them out because they still have 0 apr so no interest for users. Besides, several times people were scared they can't see their pool bc of that filter.

Frontend-v1-Original/components/ssLiquidityPairs/ssLiquidityPairsTable.tsx
```diff
@@ -278,63 +224,93 @@ const EnhancedTableToolbar = (props: PairsTableToolbarProps) => {
...
+          <li>
+            <input
+              type="radio"
+              id="all"
+              name="filter"
+              value="all"
+              checked={filter === "all"}
+              onChange={onChange}
+              className="peer hidden"
             />
```
oxSaturn, line 237: Maybe define an array to avoid duplicated code? E.g.,

```
{
  ['all', 'boosted', '...'].map(option => (<li><input type='radio' id={option} name='filter' value={option} checked={filter === option} ... />)
}
```

Another thing is the styles of options on mobile phone,



I see you set `flex-col` on purpose, but maybe try `flex-row` with `flex-wrap`?

oxSaturn (commented):
Just nitpicking. Otherwise looks good.

oxSaturn (approved):
Looks awesome.

PR #115, opened 2023-06-08
fix lock durations

oxSaturn:
Update lock durations as the max lock was set to 26 weeks in v3.

---

## PR-Codex overview
This PR adds new lock duration options and updates the lock amount and value calculations accordingly. It also displays the vesting structure in a more user-friendly way and adds a warning about the rounding of lock periods to the nearest Thursday.

### Detailed summary
- Adds new lock duration options: 6.5 weeks, 13 weeks, 19.5 weeks, and 26 weeks
- Updates lock amount and value calculations to use the new duration options
- Displays vesting structure in a more user-friendly way
- Adds a warning about the rounding of lock periods to the nearest Thursday

### DIFF
--- a/Frontend-v1-Original/components/ssVest/existingLock.tsx
+++ b/Frontend-v1-Original/components/ssVest/existingLock.tsx
@@ -9,7 +9,7 @@ import { GovToken, VestNFT, VeToken } from "../../stores/types/types";
 
 import classes from "./ssVest.module.css";
 import LockAmount from "./lockAmount";
-import LockDuration from "./lockDuration";
+import LockDuration, { lockOptions } from "./lockDuration";
 import VestingInfo from "./vestingInfo";
 
 export default function ExistingLock({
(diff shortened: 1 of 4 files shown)


oxSaturn:
I've made another two small changes:

1. show an alert about how we round down the lock end to the epoch beginning, I believe people might be confused when they set lock time to `2023-07-24`, only to find it's set to `2023-07-20` by contract.
2. I found myself trying to max lock nft multiple times but failed with the error `Can only increase lock duration` which means it's already max locked. Disabling the button would save users a click.

oxSaturn:
> are the halfed weeks for the sake of even four options?

Right, as we have examples below for 1, 0.75, 0.5, 0.25 veNFT. And equalizer uses them too, so I think it's not bad.

Actually I would love to allow users to only select date of thursday utc, but it makes the thing much complex. Might reconsider it again in near future.

t0rbik (approved):

PR #118, opened 2023-06-08
add bundle-analyzer

oxSaturn:
This is a pr to see if we can eliminate some large dependencies or replace them with smaller ones.

20230608xeC0l0OI@2x


oxSaturn:
An example of replacing `moment.js` with `date-fns` for `existingLock.tsx`:

202306090d5ADuny@2x

t0rbik:
Nice. We can also remove bignumber and use native BigInt with viem helpers. Probably can live without lottie as well. Stores module will be decreased in size when switch from flux happens. All in all something to do obviously.

oxSaturn:
After lottie gets removed:

20230613NiInJz7A@2x

We can eliminate almost 100k after gzipped.

t0rbik:
Dropping some more points we can do to improve bundle size:
1. [optimise fonts](https://everythingfonts.com/subsetter)
2. [research if any applicable to us](https://www.patterns.dev/posts/loading-sequence)

PR #124, opened 2023-06-11
update reset per v3 contracts

oxSaturn:
## PR-Codex overview
This PR updates the `ssRewards`, `ssVotes`, `ssVests`, and `stableSwapStore` components. Notable changes include:
- Replacing `actionedInCurrentEpoch` with `votedInCurrentEpoch`
- Changing the reset condition
- Updating the `_checkNFTActionEpoch` function

> The following files were skipped due to too many changes: `Frontend-v1-Original/stores/stableSwapStore.ts`


Frontend-v1-Original/components/ssVests/ssVestsTable.tsx
```diff
@@ -420,15 +419,6 @@ function MyTableRow(props: {
...
         </Typography>
-        {row.reset ? (
-          <Typography
-            variant="h5"
-            className="text-xs font-extralight"
-            color="textSecondary"
-          >
-            Reset on: {new Date(Number(row.lastVoted) * 1000).toLocaleString()}
-          </Typography>
-        ) : null}
```
oxSaturn, line 431: There's no way to track when reset happens now.

Frontend-v1-Original/stores/stableSwapStore.ts
```diff
@@ -336,16 +336,18 @@ class Store {
...
             id: tokenIndex.toString(),
             lockEnds: lockedEnd.toString(),
             lockAmount: formatUnits(lockedAmount, govToken.decimals),
             lockValue: formatUnits(lockValue, veToken.decimals),
-            actionedInCurrentEpoch,
-            reset: actionedInCurrentEpoch && !voted,
+            votedInCurrentEpoch,
+            reset:
+              (!votedInCurrentEpoch && lastVoted > 0n && voted === false) ||
+              lastVoted === 0n,
```
t0rbik, line 350: could you elaborate on the mentioned unability to get if the nft reset?

Frontend-v1-Original/stores/stableSwapStore.ts
```diff
@@ -4058,8 +4065,14 @@ class Store {
...
-            actionedInCurrentEpoch,
-            reset: actionedInCurrentEpoch && !voted,
+            votedInCurrentEpoch: votedInCurrentEpoch,
+            // all nfts are reset initially
+            // and there's one other case where reset should be true
+            // 1.1 user voted in past epoch, but not in current epoch
+            // 1.2 user called reset to set `voted` to false
+            reset:
+              (!votedInCurrentEpoch && lastVoted > 0n && voted === false) ||
+              lastVoted === 0n,
```
t0rbik, line 4075: So you cant get if reset?

oxSaturn: It depends.

1. If users call reset method on a new nft, since it just sets a `false` `voted` to `false`, we can't really tell user called reset or not from blockchain state, unless we track the event logs
2. If users call reset method on a nft voted in the past epoch but not voted in current epoch, then it would set `true` `voted` state to `false`, which means we can tell it's reset from all those conditions combined.

One different thing between this PR and v2 is the initial state of `reset`, in v2 we set it to `false` which I believe is wrong as you can merge a new nft without calling a reset, while in this pr they're initialized as `true`, might be confusing but makes more sense IMO.

Frontend-v1-Original/components/ssVests/ssVestsTable.tsx
t0rbik: I got it. so you mean we can derive if it's reseted but we can't know when it was reseted?

oxSaturn: Here's code for [reset](https://github.com/Velocimeter/contracts-latest/blob/8ba28bf61169781c4f41377254c3f2778cb72bed/contracts/Voter.sol#L113-L118) in v2:

20230611MtQ71Nkx@2x

And below is [v3](https://github.com/Velocimeter/v3-contracts/blob/01ea7d609c89b52b1eac1ff3d7075c1bca3082c9/contracts/Voter.sol#L186-L190):

202306110YUdHxWM@2x

As you can see, v3 no longer stores the timestamp to `lastVoted`.

t0rbik: i get it just double checking. initially i thought your message was saying that you can't know if nft has been reseted in v3. then I see we check for reset and i got dissonance. now i see what you meant,

PR #132, opened 2023-06-18
Feat/transfer nft

oxSaturn:
This PR is based on `feat/react-query` branch.

---

## PR-Codex overview
This PR adds the ability to transfer NFTs, improves testing, and updates dependencies.

### Detailed summary
- Added ability to transfer NFTs
- Added unit tests for `TransferNFT` component
- Updated dependencies, including `vitest` and `@vitejs/plugin-react`
- Minor changes to configuration files and code formatting

> The following files were skipped due to too many changes: `Frontend-v1-Original/yarn.lock`


Frontend-v1-Original/components/ssVest/TransferNFT.tsx
```diff
@@ -0,0 +1,121 @@
...
+        default:
+          return state;
+      }
+    },
+    {
+      address: "",
+      helperText: "",
+      error: false,
+    }
+  );
```
t0rbik, line 57: Believe it'll be much cleaner if reducer itself defined outside component

t0rbik: I'm not a fun of useReducer tbh, but it seems like it makes sense here. if you want...

Frontend-v1-Original/components/ssVest/TransferNFT.tsx
```diff
@@ -0,0 +1,121 @@
...
+            />
+          </FormControl>
+        </div>
+        <Button
+          variant="contained"
+          color="primary"
+          fullWidth
+          disabled={state.error || state.address === ""}
+          onClick={() =>
+            transferNFT(id as string, state.address as `0x${string}`)
```
t0rbik, line 113: we can add type guards for id, i see that addresses is chcked in dispatch, maybe we can set state type of address to Address?

Frontend-v1-Original/components/ssVest/TransferNFT.tsx
oxSaturn: why not? is that because of redux? lol, `reducer` here is pure, and easy to test (see commit https://github.com/Velocimeter/frontend/pull/132/commits/eb4f159b8d3a0a8a30a56d8f7bb4eabecf90a632#diff-be369d9ec48793dde76186b981d434bc801bef84305375436451fe67f7c5cc03) as long as no asynchronous states involved, when there're multiple states, it's better than `useState` IMO cause you can collocate multiple states together instead of calling multiple `setState` like:

```
setAddress();
setError();
setHelperText();
```

Frontend-v1-Original/package.json
```diff
@@ -22,6 +23,7 @@
     "@rainbow-me/rainbowkit": "^1",
     "@tanstack/react-query": "^4.29.5",
     "@tanstack/react-query-devtools": "^4.29.6",
+    "@vitejs/plugin-react": "^4.0.1",
```
t0rbik, line 26: this can go to dev dependency as well

Frontend-v1-Original/components/ssVest/TransferNFT.tsx
```diff
@@ -116,9 +116,15 @@ export function TransferNFT() {
           color="primary"
           fullWidth
           disabled={state.error || state.address === ""}
-          onClick={() =>
-            transferNFT(id as string, state.address as `0x${string}`)
-          }
+          onClick={() => {
+            if (id) {
+              if (Array.isArray(id)) {
```
t0rbik, line 121: I would think that if it's an array, then something is wrong?

Frontend-v1-Original/components/ssVest/TransferNFT.tsx
t0rbik: again, i dont have any objection. imo it's much less readable than 3 setStates in your example.
Moreover, one can argue necessity of reducer for one action type. All the same could be done in input change handler.

Frontend-v1-Original/components/ssVest/TransferNFT.tsx
oxSaturn: It really shouldn't be an array, that's why I use `as` before since I'm quite confident it should only be a string here, just like the `state.address` should only be of type `Address` when submitted because the code won't allow a invalid address to submit.

Guess it's due to next.js can't infer the type as `string` even though we were using `[id]` instead of `[...id]`.

Frontend-v1-Original/package.json
oxSaturn: My mistake.

Frontend-v1-Original/components/ssVest/TransferNFT.tsx
t0rbik: I mean maybe return void instead of passing first element of array is more logical coz we know it's not gonna be array. anyway, either way works

Frontend-v1-Original/components/ssVest/TransferNFT.tsx
oxSaturn: Just much harder to write tests when using `useState`, we're testing pure JavaScript when using `reducer`.

t0rbik (approved):

PR #136, opened 2023-06-21
optimize approve button

oxSaturn:
Currently both shows `Approve`, could be difficult to tell which is which if users don't remember at first sight.


t0rbik:
I don't think there will be to approve buttons at all in fact

oxSaturn:
I see the project is still in wip, so closing this now.

PR #142, opened 2023-06-23
update UI

oxSaturn:
Made some UI updates, thought it can improve the ux.

image

There're more I reckon can be improved, but think might better keeping the PR small, otherwise I might cause many conflicts to fix.

---

## PR-Codex overview
This PR adds new features and improvements to the frontend of the application, including the ability to redeem options and improved UI. 

### Detailed summary
- Added `cyan` color to `tailwind.config.js`
- Added `prettier` to `lint-staged`
- Improved `convert.tsx` and `redeem.tsx` components with better UI and functionality
- Added new hooks and imports to `redeem.tsx`
- Replaced hardcoded `TABS` object with dynamic `useState` in `redeem.tsx`
- Added new `lpDiscount` state and `useDiscountsData` hook to `redeemLP` component
- Improved UI and functionality of `redeemLP` component
- Improved UI of `redeemVest` component
- Other minor improvements and bug fixes

> The following files were skipped due to too many changes: `Frontend-v1-Original/components/options/redeem.tsx`

### DIFF
--- a/Frontend-v1-Original/components/options/redeem.tsx
+++ b/Frontend-v1-Original/components/options/redeem.tsx
@@ -251,7 +251,7 @@ function RedeemLiquid({ now }: { now: number }) {
       <div className="my-5 flex flex-col gap-3">
         <div className="flex items-center justify-between">
           <div
-            data-content={optionTokenSymbol}
+            data-content={`${optionTokenSymbol} to redeem`}
             className={`relative w-full border border-[rgb(46,45,45)] after:absolute after:transition-opacity ${
               areInputsEmpty ? "after:opacity-0" : "after:opacity-100"
             } after:right-2 after:top-2 after:text-xs after:text-secondary after:content-[attr(data-content)]`}
@@ -264,7 +264,7 @@ function RedeemLiquid({ now }: { now: number }) {
                   ? "text-error focus:outline-error focus-visible:outline-error"
                   : "focus:outline-secondary focus-visible:outline-secondary"
               }`}
-              placeholder={`0.00 ${optionTokenSymbol}`}
+              placeholder={`You redeem 0.00 ${optionTokenSymbol}`}
             />
           </div>
           <button className="p-4" onClick={() => setMax(INPUT.OPTION)}>
@@ -273,7 +273,7 @@ function RedeemLiquid({ now }: { now: number }) {
         </div>
         <div className="flex items-center justify-between">
           <div
-            data-content={paymentTokenSymbol}
+            data-content={`${paymentTokenSymbol} to pay`}
             className={`relative w-full border border-[rgb(46,45,45)] after:absolute after:transition-opacity ${
               areInputsEmpty ? "after:opacity-0" : "after:opacity-100"
             } after:right-2 after:top-2 after:text-xs after:text-secondary after:content-[attr(data-content)]`}
@@ -288,7 +288,7 @@ function RedeemLiquid({ now }: { now: number }) {
                   ? "text-error focus:outline-error focus-visible:outline-error"
                   : "focus:outline-secondary focus-visible:outline-secondary"
               }`}
-              placeholder={`0.00 ${paymentTokenSymbol}`}
+              placeholder={`You pay 0.00 ${paymentTokenSymbol}`}
             />
           </div>
           <button className="p-4" onClick={() => setMax(INPUT.PAYMENT)}>
@@ -296,7 +296,7 @@ function RedeemLiquid({ now }: { now: number }) {
           </button>
         </div>
         <div
-          data-content={underlyingTokenSymbol}
+          data-content={`${underlyingTokenSymbol} to receive`}
           className={`relative w-full border border-[rgb(46,45,45)] after:absolute after:transition-opacity ${
             areInputsEmpty ? "after:opacity-0" : "after:opacity-100"
           } after:right-2 after:top-2 after:text-xs after:text-secondary after:content-[attr(data-content)]`}
@@ -305,7 +305,7 @@ function RedeemLiquid({ now }: { now: number }) {
             readOnly
             value={option}
             className="w-full border-none bg-transparent p-4 text-left text-base focus:outline focus:outline-1 focus:outline-secondary focus-visible:outline-secondary"
-            placeholder={`0.00 ${underlyingTokenSymbol}`}
+            placeholder={`You get 0.00 ${underlyingTokenSymbol}`}
           />
         </div>
         {chain?.unsupported ? (
@@ -555,7 +555,7 @@ function RedeemLP({ now }: { now: number }) {
       <div className="my-5 flex flex-col gap-3">
         <div className="flex items-center justify-between">
           <div
-            data-content={optionTokenSymbol}
+            data-content={`${optionTokenSymbol} to redeem`}
             className={`relative w-full border border-[rgb(46,45,45)] after:absolute after:transition-opacity ${
               areInputsEmpty ? "after:opacity-0" : "after:opacity-100"
             } after:right-2 after:top-2 after:text-xs after:text-secondary after:content-[attr(data-content)]`}
@@ -568,7 +568,7 @@ function RedeemLP({ now }: { now: number }) {
                   ? "text-error focus:outline-error focus-visible:outline-error"
                   : "focus:outline-secondary focus-visible:outline-secondary"
               }`}
-              placeholder={`0.00 ${optionTokenSymbol}`}
+              placeholder={`You redeem 0.00 ${optionTokenSymbol}`}
             />
           </div>
           <button className="p-4" onClick={() => setMax(INPUT.OPTION)}>
@@ -577,7 +577,7 @@ function RedeemLP({ now }: { now: number }) {
         </div>
         <div className="flex items-center justify-between">
           <div
-            data-content={paymentTokenSymbol}
+            data-content={`${paymentTokenSymbol} to pay`}
             className={`relative w-full border border-[rgb(46,45,45)] after:absolute after:transition-opacity ${
               areInputsEmpty ? "after:opacity-0" : "after:opacity-100"
             } after:right-2 after:top-2 after:text-xs after:text-secondary after:content-[attr(data-content)]`}
@@ -592,12 +592,12 @@ function RedeemLP({ now }: { now: number }) {
                   ? "text-error focus:outline-error focus-visible:outline-error"
                   : "focus:outline-secondary focus-visible:outline-secondary"
               }`}
-              placeholder={`0.00 ${paymentTokenSymbol}`}
+              placeholder={`You pay 0.00 ${paymentTokenSymbol}`}
             />
           </div>
         </div>
         <div
-          data-content={underlyingTokenSymbol}
+          data-content={`${underlyingTokenSymbol} to receive`}
           className={`relative w-full border border-[rgb(46,45,45)] after:absolute after:transition-opacity ${
             areInputsEmpty ? "after:opacity-0" : "after:opacity-100"
           } after:right-2 after:top-2 after:text-xs after:text-secondary after:content-[attr(data-content)]`}
@@ -606,7 +606,7 @@ function RedeemLP({ now }: { now: number }) {
             readOnly
             value={option}
             className="w-full border-none bg-transparent p-4 text-left text-base focus:outline focus:outline-1 focus:outline-secondary focus-visible:outline-secondary"
-            placeholder={`0.00 ${underlyingTokenSymbol}`}
+            placeholder={`You get 0.00 ${underlyingTokenSymbol}`}
           />
         </div>
         {chain?.unsupported ? (
@@ -875,7 +875,7 @@ function RedeemVest({ now }: { now: number }) {
       <div className="my-5 flex flex-col gap-3">
         <div className="flex items-center justify-between">
           <div
-            data-content={optionTokenSymbol}
+            data-content={`${optionTokenSymbol} to redeem`}
             className={`relative w-full border border-[rgb(46,45,45)] after:absolute after:transition-opacity ${
               areInputsEmpty ? "after:opacity-0" : "after:opacity-100"
             } after:right-2 after:top-2 after:text-xs after:text-secondary after:content-[attr(data-content)]`}
@@ -888,7 +888,7 @@ function RedeemVest({ now }: { now: number }) {
                   ? "text-error focus:outline-error focus-visible:outline-error"
                   : "focus:outline-secondary focus-visible:outline-secondary"
               }`}
-              placeholder={`0.00 ${optionTokenSymbol}`}
+              placeholder={`You redeem 0.00 ${optionTokenSymbol}`}
             />
           </div>
           <button className="p-4" onClick={() => setMax(INPUT.OPTION)}>
@@ -897,7 +897,7 @@ function RedeemVest({ now }: { now: number }) {
         </div>
         <div className="flex items-center justify-between">
           <div
-            data-content={paymentTokenSymbol}
+            data-content={`${paymentTokenSymbol} to pay`}
             className={`relative w-full border border-[rgb(46,45,45)] after:absolute after:transition-opacity ${
               areInputsEmpty ? "after:opacity-0" : "after:opacity-100"
             } after:right-2 after:top-2 after:text-xs after:text-secondary after:content-[attr(data-content)]`}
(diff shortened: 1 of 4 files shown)


Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -46,13 +50,24 @@ import {
...
+  });
+  const { data: underlyingTokenSymbol } = useErc20Symbol({
+    address: underlyingTokenAddress,
+    enabled: !!underlyingTokenAddress,
+  });
+
+  const TABS = {
+    LP: `${paymentTokenSymbol}/${underlyingTokenSymbol} LP`,
+    VEST: `ve${underlyingTokenSymbol}`,
+    LIQUID: `${underlyingTokenSymbol}`,
```
oxSaturn, line 69: When we say `LP`, many users might not get it, what LP, which pair? there's no text telling it directly, you have to guess it from the UI.

`Vest` would lock the oToken into veToken, so maybe just show the veToken.

Same for `Liquid`, since users get liquid token like `flow`, we show what they get in the end.

Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -81,25 +96,19 @@ export function Redeem() {
...
-          value={TABS.LIQUID}
-        >
-          Redeem Liquid
-        </Tabs.Trigger>
+        {
+          Object.keys(TABS).map((tab) => (
+            <Tabs.Trigger
+              key={tab}
+              value={TABS[tab as keyof typeof TABS]}
+              className="flex h-[45px] flex-1 cursor-pointer select-none items-center justify-center bg-primaryBg px-5 text-sm leading-none text-secondary outline-none hover:text-violet-100 data-[state=active]:text-violet-100 data-[state=active]:relative data-[state=active]:shadow-[0_0_0_2px] data-[state=active]:shadow-black data-[state=active]:text-cyan"
```
oxSaturn, line 106: See https://tailwindcss.com/docs/hover-focus-and-other-states#data-attributes for `data-[]` usage.

Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -252,18 +261,16 @@ function RedeemLiquid({ now }: { now: number }) {
         <div className="flex items-center justify-between">
           <div
             data-content={optionTokenSymbol}
-            className={`relative w-full border border-[rgb(46,45,45)] after:absolute after:transition-opacity ${
-              areInputsEmpty ? "after:opacity-0" : "after:opacity-100"
-            } after:right-2 after:top-2 after:text-xs after:text-secondary after:content-[attr(data-content)]`}
+            className={`relative w-full border border-[rgb(46,45,45)] after:absolute after:transition-opacity ${areInputsEmpty ? "after:opacity-0" : "after:opacity-100"
```
oxSaturn, line 264: those are format change only.

Frontend-v1-Original/tailwind.config.js
```diff
@@ -54,6 +54,9 @@ module.exports = {
         success: "#07bc0c",
         warning: "#ffda58",
         error: "#ff4141",
+        cyan: {
+          DEFAULT: '#00E8CA',
```
oxSaturn, line 58: this color was extracted from images of https://docs.velocimeter.xyz/pressKit

Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -81,25 +96,19 @@ export function Redeem() {
       value={tab}
       onValueChange={handleTabChange}
     >
+      <h1 className="text-xl text-center mb-5">Redeem {optionTokenSymbol} Into</h1>
       <Tabs.List className="flex shrink-0" aria-label="Manage your account">
```
oxSaturn, line 100: btw, I saw you're using radix ui now. Are you going to move away from mui? From what I learned so far, radix ui is something similar to https://headlessui.com/ that focuses on the non-styles part, which means we'll have to deal with the styles ourselves then.

Frontend-v1-Original/components/options/redeem.tsx
t0rbik: elaborate?

Frontend-v1-Original/tailwind.config.js
t0rbik: what does DEFAULT do here?

Frontend-v1-Original/components/options/redeem.tsx
t0rbik: Like this change. Yes I prefer Radix and unstyled components. I would love to remove mui, but it needs time.

Frontend-v1-Original/components/options/redeem.tsx
t0rbik: All data states introduced by Radix UI https://github.com/ecklf/tailwindcss-radix

Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -46,13 +50,24 @@ import {
...
+  const { data: paymentTokenAddress } = useOptionTokenPaymentToken();
+  const { data: underlyingTokenAddress } = useOptionTokenUnderlyingToken();
+  const { data: paymentTokenSymbol } = useErc20Symbol({
+    address: paymentTokenAddress,
+    enabled: !!paymentTokenAddress,
+  });
+  const { data: underlyingTokenSymbol } = useErc20Symbol({
+    address: underlyingTokenAddress,
+    enabled: !!underlyingTokenAddress,
+  });
```
t0rbik, line 64: these symbols exist in useTokenData hook

Frontend-v1-Original/components/options/redeem.tsx
t0rbik: this looks good but dont think we should use it as state for Tabs component because it'll introduce unnecessary changes in dom. eg first render one data, then rererender after hook updates states with new data

Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -81,25 +96,19 @@ export function Redeem() {
...
-          className="flex h-[45px] flex-1 cursor-pointer select-none items-center justify-center bg-primaryBg px-5 text-[15px] leading-none text-secondary outline-none hover:text-violet-100 radix-state-active:text-violet-100 radix-state-active:focus:relative radix-state-active:focus:shadow-[0_0_0_2px] radix-state-active:focus:shadow-black"
-          value={TABS.LIQUID}
-        >
-          Redeem Liquid
-        </Tabs.Trigger>
+        {
+          Object.keys(TABS).map((tab) => (
+            <Tabs.Trigger
+              key={tab}
+              value={TABS[tab as keyof typeof TABS]}
```
t0rbik, line 105: you sure in needs type cast? does it not inferred from object keys?

Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -160,7 +169,7 @@ function RedeemLiquid({ now }: { now: number }) {
     args: [
       isValidInput(option) ? parseEther(option as `${number}`) : 0n,
       isValidInput(payment, paymentTokenDecimals) &&
-      isValidInput(maxPayment, paymentTokenDecimals)
+        isValidInput(maxPayment, paymentTokenDecimals)
```
t0rbik, line 172: ?

Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -274,20 +281,18 @@ function RedeemLiquid({ now }: { now: number }) {
...
-                  payment !== "") ||
+              className={`w-full border-none bg-transparent p-4 text-left text-base focus:outline focus:outline-1 ${(!isValidInput(payment, paymentTokenDecimals) &&
+                payment !== "") ||
                 insufficientPayment
-                  ? "text-error focus:outline-error focus-visible:outline-error"
-                  : "focus:outline-secondary focus-visible:outline-secondary"
-              }`}
+                ? "text-error focus:outline-error focus-visible:outline-error"
+                : "focus:outline-secondary focus-visible:outline-secondary"
+                }`}
```
t0rbik, line 295: ?

Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -333,13 +337,13 @@ function RedeemLiquid({ now }: { now: number }) {
...
-              isFetchingAmounts
+                waitingRedeemReceipt ||
+                writingExercise ||
+                isFetchingAmounts
                 ? "Loading..."
                 : isApprovalNeeded
-                ? "Approve"
-                : "Redeem"}
+                  ? "Approve"
+                  : "Redeem"}
```
t0rbik, line 346: ?

Frontend-v1-Original/components/options/redeem.tsx
oxSaturn: I ran `format document` command on my vscode causing those changes.

Frontend-v1-Original/tailwind.config.js
oxSaturn: You can use the color like `text-cyan` instead of `text-cyan-500`.

Frontend-v1-Original/components/options/redeem.tsx
oxSaturn: Right, but I would love to use those hooks explicitly. And `useTokenData` returns more data than I need.

Frontend-v1-Original/components/options/redeem.tsx
t0rbik: Are you running it with prettier as default formatter? Do you have prettier plugin installed (or analog if there're no plugins in your ide/editor)?

Frontend-v1-Original/tailwind.config.js
t0rbik: That essentially means that using text-cyan-`x` can cause unexpected result, as tailwind palette adjusted based on default?

Frontend-v1-Original/components/options/redeem.tsx
t0rbik: You can destruct needed data and it's being used in every tab, what is the reasoning behind your explicit intention?

t0rbik:
Let's figure out why your prettier fixes something and mine not.

Frontend-v1-Original/components/options/redeem.tsx
oxSaturn: Personal preference lol, and the hooks are exported there to use. Anyway, I'll update it, code are much simpler using `useTokenData`.

Frontend-v1-Original/components/options/redeem.tsx
oxSaturn: Added `prettier --write` to pre-commit to prevent those changes.

Frontend-v1-Original/components/options/redeem.tsx
oxSaturn: Nah, `tab` is of type `string` here, https://www.typescriptlang.org/play?#code/MYewdgzgLgBFCGAjCMC8MDeAoGubwC4YBGAJgGYAaHPRIsqm3YeirAX3xVEigG4sWAPKIAVgFNgUAHQBrcQE8IACgTIAlNIC28AA6qkaAHyYmMDuqA

image

Frontend-v1-Original/tailwind.config.js
oxSaturn: I dont' think so, it's just a shorter way to define a color https://v2.tailwindcss.com/docs/configuring-variants#default, doesn't mean tailwind will adjust the palette based on what we define. The colors tailwind provides are pre-defined.

Frontend-v1-Original/components/options/redeem.tsx
oxSaturn: Nah, it's not a state. Just a variable. So it wont cause rerender when data changes. Actually I think I should use state when possible, using variable inside component might cause stale UI problem.

Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -251,7 +251,7 @@ function RedeemLiquid({ now }: { now: number }) {
       <div className="my-5 flex flex-col gap-3">
         <div className="flex items-center justify-between">
           <div
-            data-content={optionTokenSymbol}
+            data-content={`${optionTokenSymbol} to redeem`}
```
oxSaturn, line 254: I know those changes are sort of verbose, but we have no labels for those inputs, verbose texts could be helpful in those cases:

image

image

Let me know what you think.

Frontend-v1-Original/tailwind.config.js
t0rbik: Hm, this docs are outdated and cant seem to find it in latest. Does it mean that we can use `text-cyan-lighter`, `text-cyan-dark` if we set default? Trying to understand the difference between just setting `cyan` here and defining object with `DEFAULT`.

Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -304,7 +304,7 @@ function RedeemLiquid({ now }: { now: number }) {
           <input
             readOnly
             value={option}
-            className="w-full border-none bg-transparent p-4 text-left text-base focus:outline focus:outline-1 focus:outline-secondary focus-visible:outline-secondary"
+            className="w-full border-none bg-transparent p-4 text-left text-base read-only:cursor-not-allowed focus:outline focus:outline-1 focus:outline-secondary focus-visible:outline-secondary"
```
oxSaturn, line 307: There're two types of input, one is `readOnly`, but we can't really tell it at first sight, `cursor-not-allowed` might improve the situation a little until we come up with a better solution.

Frontend-v1-Original/tailwind.config.js
oxSaturn: Check this https://github.com/tailwindlabs/tailwindcss.com/blob/d78584431db5454ed27df395f6fd6ed5fa151a2c/src/pages/docs/theme.mdx#L190, it might help you understand the use case.

In the link I provides above, we'll define `borderRadius`:

```
{
  borderRadius: {
   xl: '',
   DEFAULT: ''
  }
}
```
Then we can use `rounded` and `rounded-xl`. Just a short cut IMO.

Frontend-v1-Original/components/options/redeem.tsx
t0rbik:
```
{Object.values(TABS).map((tab) => (
          <Tabs.Trigger
            key={tab}
            value={tab}
            className="..."
          >
            {tab}
          </Tabs.Trigger>
        ))}
```
no type cast. wyt

Frontend-v1-Original/components/options/redeem.tsx
t0rbik: looks fine

Frontend-v1-Original/components/options/redeem.tsx
t0rbik: cool!

Frontend-v1-Original/components/options/redeem.tsx
t0rbik: So what's the reasoning why switch from `radix-[]` classes to `data-[]` usage?

Frontend-v1-Original/tailwind.config.js
oxSaturn: If you set `cyan` directly:

```
cyan: '#xxx'
```
It will override all `cyan` colors, `text-cyan-500`, `bg-cyan-500` etc. won't work any more.

Basically it's like calling `Object.assign({cyan: {/* some pre-defined colors*/}, {cyan: '#xxx'}})` then.

t0rbik: ok it makes sense now. thank you. so what we should do is either 1. introduce another name convention, 2. introduce whole new pallete based on changed default colour, i believe

Frontend-v1-Original/components/options/redeem.tsx
oxSaturn: smart!

Frontend-v1-Original/components/options/redeem.tsx
oxSaturn: Didn't realize it came from the tailwindcss plugin when I was editing them. But I still prefer the tailwindcss built-in syntax to plugin, as less is more, lol, and why use third-party when it's natively supported. Want me to revert?

t0rbik: it's better dx because you can use tailwind css IntelliSense with it

t0rbik: that's probably the whole point of it tbh lol

oxSaturn: Nice, you convinced me, reverted!

PR #144, opened 2023-06-24
Optimize options

oxSaturn:
## PR-Codex overview
This PR updates the styling of several components and buttons, and adds a warning message for certain actions in the Redeem component.

### Detailed summary
- Updated component styles with new border and background colors
- Changed button background and hover colors
- Added warning message for certain actions in the Redeem component
- Updated Slider component to allow selecting minimum and maximum values

> The following files were skipped due to too many changes: `Frontend-v1-Original/components/options/redeem.tsx`

### DIFF
--- a/Frontend-v1-Original/components/options/convert.tsx
+++ b/Frontend-v1-Original/components/options/convert.tsx
@@ -105,7 +105,7 @@ export function Convert() {
 
   return (
     <>
-      <div className="flex w-96 min-w-[384px] flex-col border border-primary p-5 font-sono text-lime-50 md:w-[512px] md:min-w-[512px]">
+      <div className="flex w-96 min-w-[384px] flex-col rounded-md border border-cyan/50 p-5 font-sono text-lime-50 md:w-[512px] md:min-w-[512px]">
         <div className="flex items-center justify-between">
           <div>oFLOW v1 balance</div>
           <div>{formatCurrency(optionV1Balance?.formatted)}</div>
--- a/Frontend-v1-Original/components/options/convert.tsx
+++ b/Frontend-v1-Original/components/options/convert.tsx
@@ -130,7 +130,7 @@ export function Convert() {
           </div>
           {chain?.unsupported ? (
             <button
-              className="flex h-14 w-full items-center justify-center rounded border border-transparent bg-primary p-5 text-center font-medium text-black transition-colors hover:bg-secondary focus-visible:outline-secondary disabled:bg-slate-400 disabled:opacity-60"
+              className="flex h-14 w-full items-center justify-center rounded border border-transparent bg-cyan p-5 text-center font-medium text-black transition-colors hover:bg-cyan/80 focus-visible:outline-secondary disabled:bg-slate-400 disabled:opacity-60"
               onClick={() => switchNetwork?.()}
             >
               Switch to Pulse
@@ -148,7 +148,7 @@ export function Convert() {
                 (isApprovalNeeded ? !approve : !convert)
               }
               onClick={isApprovalNeeded ? () => approve?.() : () => convert?.()}
-              className="flex h-14 w-full items-center justify-center rounded border border-transparent bg-primary p-5 text-center font-medium text-black transition-colors hover:bg-secondary focus-visible:outline-secondary disabled:bg-slate-400 disabled:opacity-60"
+              className="flex h-14 w-full items-center justify-center rounded border border-transparent bg-cyan p-5 text-center font-medium text-black transition-colors hover:bg-cyan/80 focus-visible:outline-secondary disabled:bg-slate-400 disabled:opacity-60"
             >
               {writingApprove ||
               isFetchingAllowance ||
--- a/Frontend-v1-Original/components/options/lib/useAllowance.ts
+++ b/Frontend-v1-Original/components/options/lib/useAllowance.ts
@@ -29,9 +29,18 @@ export function useAllowance() {
     args: [address!, PRO_OPTIONS.oFLOW.tokenAddress],
     enabled: !!address && !!paymentTokenAddress,
     select: (allowance) => {
-      const formattedAllowance = formatUnits(allowance, paymentTokenDecimals);
       if (!maxPayment) return;
-      return parseFloat(formattedAllowance) < parseFloat(maxPayment);
+
+      const multiplier = 1_000_000_000;
+      // this is sort of a hack that we scale up the payment first to avoid rounding errors
+      const maxPaymentBigint =
+        (BigInt(parseFloat(payment) * multiplier) *
+          10n ** BigInt(paymentTokenDecimals) *
+          101n) /
+        100n /
+        BigInt(multiplier);
+
+      return allowance < maxPaymentBigint;
     },
   });
   const { config: approveConfig } = usePrepareErc20Approve({
--- a/Frontend-v1-Original/components/options/lib/useAllowance.ts
+++ b/Frontend-v1-Original/components/options/lib/useAllowance.ts
@@ -29,18 +29,9 @@ export function useAllowance() {
     args: [address!, PRO_OPTIONS.oFLOW.tokenAddress],
     enabled: !!address && !!paymentTokenAddress,
     select: (allowance) => {
+      const formattedAllowance = formatUnits(allowance, paymentTokenDecimals);
       if (!maxPayment) return;
-
-      const multiplier = 1_000_000_000;
-      // this is sort of a hack that we scale up the payment first to avoid rounding errors
-      const maxPaymentBigint =
-        (BigInt(parseFloat(payment) * multiplier) *
-          10n ** BigInt(paymentTokenDecimals) *
-          101n) /
-        100n /
-        BigInt(multiplier);
-
-      return allowance < maxPaymentBigint;
+      return parseFloat(formattedAllowance) < parseFloat(maxPayment);
     },
   });
   const { config: approveConfig } = usePrepareErc20Approve({
--- a/Frontend-v1-Original/components/options/lib/useAllowance.ts
+++ b/Frontend-v1-Original/components/options/lib/useAllowance.ts
@@ -29,9 +29,8 @@ export function useAllowance() {
     args: [address!, PRO_OPTIONS.oFLOW.tokenAddress],
     enabled: !!address && !!paymentTokenAddress,
     select: (allowance) => {
-      const formattedAllowance = formatUnits(allowance, paymentTokenDecimals);
       if (!maxPayment) return;
-      return parseFloat(formattedAllowance) < parseFloat(maxPayment);
+      return allowance < parseUnits(maxPayment, paymentTokenDecimals);
     },
   });
   const { config: approveConfig } = usePrepareErc20Approve({
--- a/Frontend-v1-Original/components/options/redeem.tsx
+++ b/Frontend-v1-Original/components/options/redeem.tsx
@@ -616,7 +616,7 @@ function RedeemLP({ now }: { now: number }) {
           />
         </div>
         {isSelectedDurationLessThanLockEnd && (
-          <div className="flex flex-col items-start justify-center text-warning">
+          <div className="flex flex-col items-start justify-center space-y-2 text-warning">
             <div className="uppercase">WARNING</div>
             <div className="text-sm">
               You already have a staked position expires on{" "}
@@ -626,7 +626,7 @@ function RedeemLP({ now }: { now: number }) {
           </div>
         )}
         {isSelectedDurationMoreThanLockEnd && (
-          <div className="flex flex-col items-start justify-center text-warning">
+          <div className="flex flex-col items-start justify-center space-y-2 text-warning">
             <div className="uppercase">WARNING</div>
             <div className="text-sm">
               You are going to increase your lock end from{" "}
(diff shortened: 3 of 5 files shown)


Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -390,7 +390,13 @@ function RedeemLP({ now }: { now: number }) {
 
   const { maxLpDiscount, minLpDiscount } = useDiscountsData();
 
-  const [lpDiscount, setLpDiscount] = useState(maxLpDiscount ?? 50);
```
oxSaturn, line 393: Obviously I was wrong to assume `maxLpDiscount` to be available when calling `useState`. It will only be `undefined` when calling `useState`. Hence we'll rely `useEffect` to update the value.

Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -390,7 +390,13 @@ function RedeemLP({ now }: { now: number }) {
...
+  const [lpDiscount, setLpDiscount] = useState(100);
```
oxSaturn, line 393: I'm using `100` instead of `50` here due to the rerender caused by `useEffect` when we switch to lp tab from other tabs, i.e., the slider would move from `50` to `maxLpDiscount`.


https://github.com/Velocimeter/frontend/assets/126733611/58371190-42ad-4025-b916-fbe86d8509f2



With `lpDiscount` set to `100`, there won't be such a problem because the max value of slider is `maxLpDiscount`.

Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -547,13 +547,17 @@ function RedeemLP({ now }: { now: number }) {
...
       </div>
-      <Slider
-        value={[lpDiscount]}
-        onValueChange={(e) => setLpDiscount(e[0])}
-        min={minLpDiscount ?? 0}
-        max={maxLpDiscount ?? 100}
-        className="relative flex h-5 touch-none select-none items-center"
-      />
+      <div className="flex items-center space-x-2">
+        <span>{minLpDiscount}%(min)</span>
```
oxSaturn, line 551: Currently I have to drag the slider to both sides to find the min discount and max discount, thought it might not be a bad idea to show them right on the UI.

Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -615,69 +615,35 @@ function RedeemLP({ now }: { now: number }) {
             placeholder={`You get 0.00 ${underlyingTokenSymbol}`}
           />
         </div>
-        {chain?.unsupported ? (
```
oxSaturn, line 608: When I was using the page on laptop, I can only saw the `Warning` heading in the first screen:

image

Since users should read the content of warning first before any other actions, I think it would be more reasonable to move the warning section above the button:

image

Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -615,69 +615,35 @@ function RedeemLP({ now }: { now: number }) {
...
           <div className="flex flex-col items-start justify-center text-warning">
-            <div className="underline">WARNING</div>
-            <div>
-              You already have a staked position expires{" "}
-              {dayjs.unix(stakedLockEnd).fromNow()}. You can only redeem with
-              same lock or longer.
+            <div className="uppercase">WARNING</div>
+            <div className="text-sm">
+              You already have a staked position expires on{" "}
+              {dayjs.unix(stakedLockEnd).format("YYYY-MM-DD HH[:]mm")}. You can
```
oxSaturn, line 623: Besides, I'm changing the date format here, more specific. Personally, I don't like ambiguous date like `in a month`, especially in defi. Users should know the exact date for unlocking so they can plan for it.

Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -615,69 +615,35 @@ function RedeemLP({ now }: { now: number }) {
...
+                  .format("YYYY-MM-DD HH[:]mm")}
               </span>
               .
             </div>
-            <Separator.Root className="my-[15px] bg-warning radix-orientation-horizontal:h-px radix-orientation-horizontal:w-full" />
-            <div className="flex items-center">
+            <div className="my-3 flex items-center">
               <Checkbox.Root
-                className="flex h-[25px] w-[25px] appearance-none items-center justify-center rounded-[4px] bg-secondary outline-none"
+                className="flex h-[25px] w-[25px] appearance-none items-center justify-center rounded-[4px] border-2 border-solid border-secondary outline-none radix-state-checked:bg-secondary"
```
oxSaturn, line 646: Made a small change for unchecked status:

Before:

image

After:

image

Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -615,69 +615,35 @@ function RedeemLP({ now }: { now: number }) {
...
+            <div className="text-sm">
+              You already have a staked position expires on{" "}
+              {dayjs.unix(stakedLockEnd).format("YYYY-MM-DD HH[:]mm")}. You can
+              only redeem with same lock or longer.
             </div>
           </div>
         )}
         {isSelectedDurationMoreThanLockEnd && (
           <div className="flex flex-col items-start justify-center text-warning">
-            <div className="underline">WARNING</div>
```
oxSaturn, line 655: `underline` was removed as `underline` is used for links most of the time. `uppercase` is enough to distinguish the header from body especially when we have body size changed to `text-sm`.

Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -698,49 +698,51 @@ function RedeemLP({ now }: { now: number }) {
...
-        <div className="flex flex-col items-start justify-center text-sm">
-          <div className="underline">Breakdown</div>
-          <div>
+        <div className="h-1" />
+        <div className="flex flex-col items-start justify-center space-y-2">
+          <div className="uppercase">Breakdown</div>
+          <div className="text-sm">
             You get {lpDiscount}% discount. There is additional token transfer
             for LP creation.
           </div>
```
oxSaturn, line 707: From what I can tell, contents under `Breakdown` are a cohesive section, hence I'm restructuring the HTML and make some changes to the styles.

image

Before:

image

Frontend-v1-Original/components/options/convert.tsx
```diff
@@ -105,7 +105,7 @@ export function Convert() {
 
   return (
     <>
-      <div className="flex w-96 min-w-[384px] flex-col border border-primary p-5 font-sono text-lime-50 md:w-[512px] md:min-w-[512px]">
+      <div className="flex w-96 min-w-[384px] flex-col rounded-md border border-cyan/50 p-5 font-sono text-lime-50 md:w-[512px] md:min-w-[512px]">
```
oxSaturn, line 108: Tried to use more color from our brand.

Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -694,7 +694,7 @@ function RedeemLP({ now }: { now: number }) {
             isFetchingAllowanceOrApproving
               ? "Loading..."
               : isApprovalNeeded
-              ? "Approve"
+              ? `Approve ${paymentTokenSymbol}`
```
oxSaturn, line 697: Don't know what I'm approving here when I see the button, `oFlow`, or `usdc`, or something else? Yes, the wallet will tell us when button clicked, but thought it would be much better if we can show it right on the button.

oxSaturn:
I can see a bug that after approved, the approve button won't turn into `Redeem into usdc/flow lp` button, I had to remove some decimals from `oFlow` token to make it work. There was a similar report in discord https://discord.com/channels/1060312715496783952/1106906000084172870/1121844563242393801

Here's the log result of `useAllownance`:

image

image

Frontend-v1-Original/components/options/lib/useAllowance.ts
```diff
@@ -29,9 +29,18 @@ export function useAllowance() {
...
+      const multiplier = 1_000_000_000;
+      // this is sort of a hack that we scale up the payment first to avoid rounding errors
+      const maxPaymentBigint =
+        (BigInt(parseFloat(payment) * multiplier) *
+          10n ** BigInt(paymentTokenDecimals) *
+          101n) /
+        100n /
+        BigInt(multiplier);
+
+      return allowance < maxPaymentBigint;
```
oxSaturn, line 43: Fix bug reported here https://github.com/Velocimeter/frontend/pull/144#issuecomment-1605429736.

Generally I don't think it's not a good idea to deal with decimals when we do math in JS. We'd better do math in BigInt, and only convert them to Number when presenting them on the UI.

Frontend-v1-Original/components/options/lib/useAllowance.ts
```diff
@@ -29,9 +29,8 @@ export function useAllowance() {
     args: [address!, PRO_OPTIONS.oFLOW.tokenAddress],
     enabled: !!address && !!paymentTokenAddress,
     select: (allowance) => {
-      const formattedAllowance = formatUnits(allowance, paymentTokenDecimals);
       if (!maxPayment) return;
-      return parseFloat(formattedAllowance) < parseFloat(maxPayment);
+      return allowance < parseUnits(maxPayment, paymentTokenDecimals);
```
oxSaturn, line 33: Closes https://github.com/Velocimeter/frontend/pull/144#issuecomment-1605429736

The problem here is when we approve, the number will be processed by `parseUnits` first:

image

I.e., numbers like `4.3456781` will be rounded down into `4345678n`.

While we don't process the `maxPayment` number first, hence 4.345678 < 4.3456781, stuck.

Frontend-v1-Original/components/options/redeem.tsx
oxSaturn: I can sort of understand why you place warning under the button, is that because of the flicker caused by show/hide of warning when we drag the slider?

Frontend-v1-Original/components/options/redeem.tsx
t0rbik: the other way is to add setter back to default value in tab switch handler

Frontend-v1-Original/components/options/redeem.tsx
t0rbik: idk, feels like it redundant. especially saying (min) and (max) in braces, like it's pretty obvious. Plus one still would be curious to go from min to max to see lock times accordingly.

Frontend-v1-Original/components/options/redeem.tsx
t0rbik: Exactly

Frontend-v1-Original/components/options/redeem.tsx
t0rbik: cool

Frontend-v1-Original/components/options/convert.tsx
t0rbik: you're right that we should round borders, although i like square more, but yeh

Frontend-v1-Original/components/options/redeem.tsx
oxSaturn: Agreed, min/max are redundant, will remove them. But would still love to show `minLpDiscount` and `maxLpDiscount` on slider's sides so I can tell the discount range right from the UI without dragging.

Frontend-v1-Original/components/options/redeem.tsx
oxSaturn: Will keep these changes now that you've fixed the flickering problem.

Frontend-v1-Original/components/options/redeem.tsx
```diff
@@ -86,7 +86,7 @@ export function Redeem() {
       <h1 className="mb-5 text-center text-xl">
         Redeem {optionTokenSymbol} Into
       </h1>
-      <Tabs.List className="flex shrink-0" aria-label="Manage your account">
+      <Tabs.List className="flex shrink-0" aria-label={`Redeem options`}>
```
t0rbik, line 89: nice find. but why not simple string?

oxSaturn: lol, I should introduce a eslint rule or whatever to enforce those conventions.

Frontend-v1-Original/components/options/convert.tsx
```diff
@@ -130,7 +130,7 @@ export function Convert() {
           </div>
           {chain?.unsupported ? (
             <button
-              className="flex h-14 w-full items-center justify-center rounded border border-transparent bg-primary p-5 text-center font-medium text-black transition-colors hover:bg-secondary focus-visible:outline-secondary disabled:bg-slate-400 disabled:opacity-60"
+              className="flex h-14 w-full items-center justify-center rounded border border-transparent bg-cyan p-5 text-center font-medium text-black transition-colors hover:bg-cyan/80 focus-visible:outline-secondary disabled:bg-slate-400 disabled:opacity-60"
```
oxSaturn, line 133: This is another chance to use `cyan`, but I'm not against using canto green here if you prefer.

But there was an issue here, i.e., disabled button shares a similar background with a clickable button when hovered.

image

oxSaturn:
Think I won't push new commits to this PR then, will leave any of them to other PRs.

Frontend-v1-Original/components/options/convert.tsx
oxSaturn: Here's what it looks like after this commit:

image

Left is clickable button, right one is when you hover your mouse over a clickable button.

t0rbik (approved):

PR #152, opened 2023-06-26
show usd value of staked in gauge

t0rbik:
## PR-Codex overview
This PR adds two new variables to the `Stake` component in `stakedWithoutLockValue` and `stakedWithLockValue`. It also updates the `useStakeData` hook to include these new variables in the `useTotalStaked` function.

### Detailed summary
- Adds `stakedWithoutLockValue` and `stakedWithLockValue` variables to `Stake` component
- Updates `useStakeData` hook to include new variables in `useTotalStaked` function

### DIFF
--- a/Frontend-v1-Original/components/options/lib/useStakeData.ts
+++ b/Frontend-v1-Original/components/options/lib/useStakeData.ts
@@ -63,17 +63,22 @@ export function useStakeData() {
 
   const stakedBalanceWithoutLock =
     stakedBalance && stakedBalanceWithLock
-      ? parseFloat(stakedBalance) - parseFloat(stakedBalanceWithLock)
+      ? (
+          parseFloat(stakedBalance) - parseFloat(stakedBalanceWithLock)
+        ).toString()
       : undefined;
 
-  const { data: gaugeTotalStakedData } = useTotalStaked();
+  const { data: gaugeTotalStakedData } = useTotalStaked(
+    stakedBalanceWithoutLock,
+    stakedBalanceWithLock
+  );
 
   return {
     ...gaugeTotalStakedData,
     pair,
     pooledBalance,
     stakedBalance,
-    stakedBalanceWithoutLock: stakedBalanceWithoutLock?.toString(),
+    stakedBalanceWithoutLock,
     stakedBalanceWithLock,
     stakedLockEnd,
     paymentTokenBalanceToDistribute:
@@ -89,7 +94,10 @@ export function useStakeData() {
   };
 }
 
-export function useTotalStaked() {
+export function useTotalStaked(
+  stakedBalanceWithoutLock: string | undefined,
+  stakedBalanceWithLock: string | undefined
+) {
   const { data: pair } = useMaxxingGaugeStake();
   const { data: totalSupply } = useMaxxingGaugeTotalSupply({
     select: (data) => formatEther(data),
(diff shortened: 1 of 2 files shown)


Frontend-v1-Original/components/options/lib/useStakeData.ts
```diff
@@ -98,20 +106,49 @@ export function useTotalStaked() {
...
+        stakedBalanceWithoutLock,
+        stakedBalanceWithLock
+      ),
+    enabled:
+      !!pair &&
+      !!totalSupply &&
+      !!pairData &&
+      !!tokenPrices &&
+      !!stakedBalanceWithoutLock &&
+      !!stakedBalanceWithLock,
```
oxSaturn, line 132: From what I know, the `queryFn` here will only run when `enabled` is true, so do we really still need to throw an error?

Frontend-v1-Original/components/options/lib/useStakeData.ts
```diff
@@ -98,20 +106,49 @@ export function useTotalStaked() {
...
-  if (!pairData || !totalSupply) {
-    throw new Error("Pair data or total supply is undefined");
+  if (
+    !pairData ||
+    !totalSupply ||
+    !tokenPrices ||
+    !stakedWithoutLock ||
+    !stakedWithLock
+  ) {
+    throw new Error("Total staked error");
```
oxSaturn, line 151: I mean here, `pairData` etc. were already checked in `enabled`.

Frontend-v1-Original/components/options/lib/useStakeData.ts
t0rbik: it's just for typescript in fact. otherwise you need to update function params in definiton and pass it with Non-null Assertion Operator (Postfix!).
Reference: [tkdodo](https://tkdodo.eu/blog/react-query-and-type-script#:~:text=I%27ve%20found%20the%20best%20way%20to%20go%20here%2C%20if%20you%20don%27t%20like%20the%20non%2Dnull%20assertion%20operator%2C%20is%20to%20accept%20that%20id%20can%20be%20undefined%20and%20reject%20the%20Promise%20in%20the%20queryFn.%20It%27s%20a%20bit%20of%20duplication%2C%20but%20it%27s%20also%20explicit%20and%20safe%3A)

oxSaturn (approved):

Frontend-v1-Original/components/options/lib/useStakeData.ts
oxSaturn: Ok, I think I just came up with another hack.

```
queryFn: () => getTotalStaked(pairData!, totalSupply!, tokenPrices!),
```

t0rbik: > Technically, TypeScript is right, id is possibly undefined: the enabled option does not perform any type narrowing. Also, there are ways to bypass the enabled option, for example by calling the refetch method returned from useQuery. In that case, the id might really be undefined.
> 
> I've found the best way to go here, if you don't like the [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator), is to accept that id can be undefined and reject the Promise in the queryFn. It's a bit of duplication, but it's also explicit and safe.

oxSaturn: Read that, but we don't call `refetch` in this case, do we?

t0rbik: what if we do?

t0rbik: i mean in any point in future

t0rbik: i guess i just like the way without non null assertion better

PR #154, opened 2023-06-27
show apr separately for each token

t0rbik:
## PR-Codex overview
This PR removes the `@radix-ui/react-separator` package, updates `@mui/material` to version `5.11.5`, and refactors the `useGaugeApr` function to return an array of token objects with their respective APR ranges.

### Detailed summary
- Removes `@radix-ui/react-separator` package.
- Updates `@mui/material` to version `5.11.5`.
- Refactors `useGaugeApr` function to return an array of token objects with their respective APR ranges.

### DIFF
--- a/Frontend-v1-Original/components/options/stake.tsx
+++ b/Frontend-v1-Original/components/options/stake.tsx
@@ -64,8 +64,9 @@ export function Stake() {
     const arr = [];
 
     for (const [token, aprRange] of aprMap.entries()) {
-      console.log(token, aprRange);
       if (aprRange[0] === aprRange[1]) {
+        arr.push({ ...token, apr: aprRange[0] });
+      } else if (aprRange[0] < 0.01 && aprRange[1] < 0.01) {
         arr.push({ ...token, apr: aprRange[0] });
       } else {
         arr.push({ ...token, aprRange });
--- a/Frontend-v1-Original/components/options/stake.tsx
+++ b/Frontend-v1-Original/components/options/stake.tsx
@@ -22,13 +22,7 @@ import {
   usePrepareErc20Approve,
 } from "../../lib/wagmiGen";
 
-import {
-  isValidInput,
-  useStakeData,
-  useGaugeApr,
-  useTokenData,
-  useGaugeRewardTokens,
-} from "./lib";
+import { isValidInput, useStakeData, useGaugeApr, useTokenData } from "./lib";
 
 const ACTION = {
   STAKE: "STAKE",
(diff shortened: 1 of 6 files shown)


Frontend-v1-Original/components/options/stake.tsx
```diff
@@ -61,23 +55,26 @@ export function Stake() {
...
-    for (const token of rewardTokens) {
-      if (token.logoUrl !== undefined && tokenLogosSet.has(token.logoUrl)) {
-        continue;
-      } else if (token.logoUrl !== undefined) {
-        tokenLogosSet.add(token.logoUrl);
+
+    for (const [token, aprRange] of aprMap.entries()) {
+      if (aprRange[0] === aprRange[1]) {
+        arr.push({ ...token, apr: aprRange[0] });
+      } else if (aprRange[0] < 0.01 && aprRange[1] < 0.01) {
```
t0rbik, line 69: Here we can probably check only for `aprRange[0]`

Frontend-v1-Original/components/options/stake.tsx
```diff
@@ -61,23 +55,26 @@ export function Stake() {
...
   } = useStakeData();
-  const { data: aprRange } = useGaugeApr();
-  const { data: rewardTokens } = useGaugeRewardTokens();
+
+  const { data: aprMap } = useGaugeApr();
+
   const displayedRewardTokens = useMemo(() => {
-    if (!rewardTokens) return undefined;
-    const tokenLogosSet = new Set<string>();
+    if (!aprMap) return undefined;
```
oxSaturn, line 62: I think this won't be a false value as you initialized an empty value of `new Map()`.

Frontend-v1-Original/components/options/stake.tsx
```diff
@@ -227,35 +224,40 @@ export function Stake() {
           <div>Total staked</div>
           <div>${formatCurrency(totalStakedValue)}</div>
         </div>
-        <div className="flex items-center justify-between">
-          <div className="flex items-center justify-start gap-2">
-            <div>APR</div>
-            {displayedRewardTokens && displayedRewardTokens.length > 0 && (
-              <div className="flex items-center justify-center gap-1">
-                {displayedRewardTokens.map((token) => (
+        <div className="my-1 flex items-center justify-between">
```
oxSaturn, line 227:
```suggestion
        <div className="flex items-start justify-between">
```
I would remove `my-1` here to keep consistency with others.

As for horizontal alignment, I would prefer `items-start` in this case. But it's up to you.

image

Frontend-v1-Original/components/options/stake.tsx
oxSaturn, line 69: Just curious, why not return what we need directly in the `useGaugeApr` file? So we don't have to preprocess the result here for rendering purpose. Seems feasible unless there're any problems I don't know.

Frontend-v1-Original/components/options/stake.tsx
```diff
@@ -227,35 +224,40 @@ export function Stake() {
...
+                      : `
+                  ${formatCurrency(token.apr)} %
+                  `}
+                  </div>
                   <img
-                    key={token.address}
                     src={token.logoUrl ?? "/tokens/unknown-logo.png"}
                     alt={`${token.symbol} logo`}
-                    className="h-5 w-5 rounded-full"
+                    className="block h-5 w-5 rounded-full"
```
oxSaturn, line 248:
```suggestion
                    className="block h-5 w-5 rounded-full"
                    title={token.symbol}
```
We use same icon for both flow and oFlow, it could be confusing, so maybe add a title to help?

oxSaturn (approved):

PR #159, opened 2023-06-29
make gauge reactive by reading it from option token

t0rbik:
### Detailed summary
- This PR focuses on updating the usage of the `useOptionTokenGauge` hook in multiple files.
- The `gaugeAddress` is now passed as a parameter to the `useOptionTokenGauge` hook in the following files:
  - `components/options/reward.tsx`
  - `components/options/stake.tsx`
  - `components/options/lib/useStakeData.ts`
  - `components/options/lib/useGaugeRewards.ts`
- The `useOptionTokenGauge` hook is used to retrieve the `gaugeAddress` data.
- The `gaugeAddress` parameter is used in various configurations and arguments for different functions and hooks.


oxSaturn (approved):
So the gauge address could be updated when needed, right?

t0rbik:
Yes and we don't need to keep track of it when we have many options. I think we will have a factory with option contracts where we can ask for addresses

PR #203, opened 2023-07-03
fix sticky

oxSaturn:
The sticky is broken due to the `overflow`. Removing them would fix the problem.

Personally I would rather removing `sticky` at all, as we don't have a long page to scroll, user can scroll back to navbar easily, and sticky element just take up quite a lot of space in small screens like laptop:

image

### DIFF
--- a/Frontend-v1-Original/components/header/header.tsx
+++ b/Frontend-v1-Original/components/header/header.tsx
@@ -35,7 +35,7 @@ function Header() {
   return (
     <>
       <div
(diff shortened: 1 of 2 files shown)


t0rbik:
Let's remove sticky. make it 
```
div flex flex-col
header
main flex-[1]
footer

wyt?

oxSaturn:
> Let's remove sticky. make it
> 
> ```
> div flex flex-col
> header
> main flex-[1]
> footer
> /div
> ```
> 
> wyt?

`footer` would be my next plan, let's see if we can make it this simple.

t0rbik (approved):

PR #209, opened 2023-07-03
exclude e2e tests from vitest

oxSaturn:
Sorry, just didn't notice the unit test was failing here https://github.com/Velocimeter/frontend/pull/207

### DIFF
--- a/Frontend-v1-Original/vitest.config.ts
+++ b/Frontend-v1-Original/vitest.config.ts
@@ -1,4 +1,4 @@
-import { defineConfig } from "vitest/config";
+import { configDefaults, defineConfig } from "vitest/config";
 import react from "@vitejs/plugin-react";
 
 export default defineConfig({
@@ -6,5 +6,6 @@ export default defineConfig({
   test: {
     environment: "jsdom",
     setupFiles: ["./scripts/env-setup.ts"],
+    exclude: [...configDefaults.exclude, "**/e2e/**"],
   },
 });


oxSaturn:
LOL, then we have more problems https://github.com/WalletConnect/walletconnect-monorepo/issues/683

.github/workflows/playwright.yml
```diff
@@ -16,7 +17,7 @@ jobs:
     - name: Install dependencies
       run: yarn
     - name: Install Playwright Browsers
-      run: yarn playwright install --with-deps
+      run: yarn playwright install --with-deps chromium
```
oxSaturn, line 20: Currently we only test against chrome.

Frontend-v1-Original/stores/connectors/viem.ts
```diff
@@ -73,17 +73,22 @@ const { chains, publicClient } = configureChains(
 );
 
 const projectId = process.env.NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID!;
+// we don't really have NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID under testing environment
```
t0rbik, line 76: can we just add another wallet project id to testing env?

oxSaturn: yes, we can, but is it a good idea to expose it in `.env.test`? Oh, actually we can set up a environment variable to use in github action, just not worth the hustle at the moment IMO since we don't really use wallet connect in tests.

t0rbik: it's free and no real hard tbh to expose it

t0rbik: oof their dashboard is so slow

t0rbik: ef237d3be67ef40d50f2a7337945c124 you can use this

t0rbik: ok neberminde it seems to be a pain their dashboard

PR #215, opened 2023-07-04
FIX/To distribute usd value 

t0rbik:
### Detailed summary
- The PR focuses on updating the `Stake` component to display the `paymentTokenToDistributeValue` instead of `paymentTokenBalanceToDistribute`.
- It introduces a new function `usePaymentTokenToDistributeValue` to calculate the value of the payment token to be distributed.
- It adds a new helper function `getPaymentTokenToDistributeValue` to calculate the value of the payment token to be distributed based on the token prices.

### DIFF
--- a/Frontend-v1-Original/components/options/lib/useStakeData.ts
+++ b/Frontend-v1-Original/components/options/lib/useStakeData.ts
@@ -81,6 +81,15 @@ export function useStakeData() {
     stakedBalanceWithLock
   );
 
+  const paymentTokenBalanceToDistribute =
+    paymentTokenBalanceInOption && paymentTokenLeftInGauge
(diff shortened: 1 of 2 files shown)


t0rbik:
accident

t0rbik:
accidentally closed lol

oxSaturn (commented):
Would be helpful for reviewing if you can share in the description about the reason to introduce changes next time. Otherwise I had to guess.

t0rbik:
Ooooh, right. Apologies. So we're currently showing `amount of token reward` with $ symbol because on pulse it's usdc reward. That works for pulse, but on fantom payment token is WFTM. That's being said, we now show wrong data bc `amount of token reward` !== `usd value of token reward`

Issue #223, opened 2023-07-05
Vesting fvm until 2024.7.03


oxSaturn:
User ran into `voting lock can be 52 weeks max` error.

I.e., could be caused by this line of code https://github.com/Velocimeter/frontend/blob/d34b2bd4ad5ac117374634ce4527ef942760b716/Frontend-v1-Original/components/ssVest/lockDuration.tsx#L27

t0rbik:
does it need chores to pulse and canto?

oxSaturn:
> does it need chores to pulse and canto?

yep, will cherry pick it.

oxSaturn:
Fixed.

PR #225, opened 2023-07-05
fix max lock error

oxSaturn:
Closes https://github.com/Velocimeter/frontend/issues/223

Not sure why the `expiry` would add another day here, anyway, we'll compare the result with the max seconds we can use to decide the unlockTime.

---

### Detailed summary

- Added a new variable `maxLockDurationInSeconds`
- Modified the `createVest` function to consider the `maxLockDurationInSeconds` when setting the `unlockTime` property.


t0rbik (approved):
looks good lets ship it

PR #231, opened 2023-07-06
[canto-v2] Cherry pick/mobile UI canto v2

oxSaturn:
### Detailed summary
- Added `.test-results/`, `.playwright-report/`, and `.playwright/.cache/` to `.gitignore` file.
- Added `cyan` color to `tailwind.config.js`.
- Updated `layout.tsx` to remove `sticky` class from `div` element.
- Updated `vitest.config.ts` to exclude `**/e2e/**` from test coverage.
- Added `test:e2e` script to `package.json` to run Playwright tests.
- Added `e2e/navbar-clickable.test.ts` for testing navbar clickability.
- Updated `PageWrapper.tsx` to change `h-[calc(100%-150px)]` to `h-full` in `Paper` component.
- Added `playwright.yml` workflow file for running Playwright tests.
- Updated `viem.ts` to use default RPC URL if `NEXT_PUBLIC_RPC_URL` is not defined.
- Updated `header.tsx` to change `SiteLogo` dimensions.
- Updated `mobileHeader.tsx` to change `SiteLogo` dimensions.
- Updated `header.tsx` to change `Close` button dimensions.
- Updated `header.tsx` to add empty `div` elements for spacing.
- Updated `header.tsx` to add `span` element for `ConnectButton` positioning.
- Updated `playwright.config.ts` to configure Playwright tests

> The following files were skipped due to too many changes: `Frontend-v1-Original/components/header/info.tsx`, `Frontend-v1-Original/components/navigation/navigation.tsx`

### DIFF
--- a/Frontend-v1-Original/stores/connectors/viem.ts
+++ b/Frontend-v1-Original/stores/connectors/viem.ts
@@ -43,7 +43,7 @@ const { chains, publicClient } = configureChains(
   [
     jsonRpcProvider({
       rpc: () => ({
-        http: process.env.NEXT_PUBLIC_RPC_URL!,
+        http: process.env.NEXT_PUBLIC_RPC_URL ?? 'https://canto.gravitychain.io',
       }),
     }),
     jsonRpcProvider({
--- a/Frontend-v1-Original/stores/connectors/viem.ts
+++ b/Frontend-v1-Original/stores/connectors/viem.ts
@@ -43,7 +43,7 @@ const { chains, publicClient } = configureChains(
   [
     jsonRpcProvider({
       rpc: () => ({
-        http: process.env.NEXT_PUBLIC_RPC_URL!,
+        http: "https://canto.gravitychain.io",
       }),
     }),
     jsonRpcProvider({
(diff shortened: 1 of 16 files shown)


Frontend-v1-Original/stores/connectors/viem.ts
```diff
@@ -43,7 +43,7 @@ const { chains, publicClient } = configureChains(
   [
     jsonRpcProvider({
       rpc: () => ({
-        http: process.env.NEXT_PUBLIC_RPC_URL!,
+        http: "https://canto.gravitychain.io",
```
oxSaturn, line 46: Can we just use the string here? I saw both other branches uses strings instead of reading from env. And it's causing e2e test failed https://github.com/Velocimeter/frontend/actions/runs/5474664234/jobs/9969727731 due to playwright not reading `.env` files.

t0rbik: that's because we dont have any owned rpcs on other chains. as for canto it lacks a normal one and our rpc is private so we shouldn't really just display it here, though I understand that it's still publicly accessible. would be not best thing to just put it here

t0rbik (changes requested):

t0rbik (approved):

PR #253, opened 2023-07-09
make lz filter less confusing and only show complete lz assets

t0rbik:
show only lz-lz pairs. do not confuse for example that ERN is layer zero

---

### Focus of the PR:
This PR focuses on updating the condition in the `LiquidityPairsTable.tsx` file to return `false` when the filter is "layer zero" and either `pair.token0_address` or `pair.token1_address` is not included in `lzTokensLowerCase`.

### Detailed summary:
- Updated the condition to use `||` instead of `&&` in the `if` statement.


oxSaturn (approved):
Still confusing to the end users IMO as I can see other tabs containing `usdc` paired with non-L0 tokens.

And this `layer zero` tab works more like `L0 Only`.

But I can't see a good solution unless we label those with a `lz` prefix, lol.

PR #257, opened 2023-07-11
fvm staking

oxSaturn:
## PR-Codex overview
This PR focuses on adding stake functionality to the application. 

### Detailed summary
- Added stake-related components and hooks
- Updated navigation to include a "Stake" option
- Added stake-related functions and constants
- Updated the configuration file with stake-related information

> The following files were skipped due to too many changes: `components/stake/StakeFVM.tsx`, `stores/abis/stakeFvmABI.ts`


tailwind.config.js
```diff
@@ -51,6 +51,13 @@ module.exports = {
         cyan: {
           DEFAULT: "#00E8CA",
           700: "#00CFB3",
+          900: "#00695B",
+        },
+        green: {
+          900: "#021716",
```
oxSaturn, line 57: This is actually not a new value, it's same as `appBackground`, I'm defining another name to stay away from semantic naming.

components/stake/StakeFVM.tsx
```diff
@@ -0,0 +1,356 @@
...
+  }
+}
+
+function SubHeader(props: { text: string }) {
+  return <h2 className="text-2xl text-white">{props.text}</h2>;
+}
+
+function Section(props: { children: ReactNode }) {
+  return (
+    <div className="bg-green-900 space-y-3 p-5 border border-cyan-900/70 rounded-md">
```
oxSaturn, line 61: I'm applying a solid background color here because of our background image. It makes our page style more difficult to handle even though background was blurred, e.g.,

CleanShot-3vWVCaWG@2x

Solid color makes it easier to styling our elements.

components/navigation/navigation.tsx
```diff
@@ -43,6 +46,7 @@ function Navigation() {
         {renderSubNav("Vote", "vote")}
         {renderSubNav("Rewards", "rewards")}
         {renderSubNav("Options", "options")}
+        {renderSubNav("Stake", "stake")}
```
oxSaturn, line 49: lmao, we just can't stop pushing new menu into navbar.

components/stake/StakeFVM.tsx
```diff
@@ -0,0 +1,356 @@
...
+   */
+
+  const {
+    data: fvmBalance,
+    isFetching: isFetchingFvmBalance,
+    refetch: refetchFvmBalance,
+  } = useBalance({
+    address: address!,
+    token: GOV_TOKEN_ADDRESS,
+    enabled: !!GOV_TOKEN_ADDRESS && !!address,
```
oxSaturn, line 86: I'm hesitate to include `!!address` here, as We shouldn't be able to mount this component if `address` is of any false value. Remember our `PageWrapper` component, when `address` is false, it will render a placeholder component.

tailwind.config.js
```diff
@@ -51,6 +51,13 @@ module.exports = {
...
+        },
+        blue: {
+          DEFAULT: "#0C8FE8",
```
oxSaturn, line 60: yeah, a new color, but I'm definitely not a professional designer, so do not hesitate to push back if you think it's not a good idea!

components/navigation/navigation.tsx
t0rbik: lol

components/stake/StakeFVM.tsx
t0rbik: we want to eventually move from placeholder when wallet not connected, so it's nice to think before hand

oxSaturn:
There're many more other small improvements we can make, but I reckon it might not worth it. Will skip them and go ahead with the APR thing first.

components/stake/StakeFVM.tsx
oxSaturn: There is this saying that `Premature optimization is the root of all evil`, lol.

components/stake/StakeFVM.tsx
```diff
@@ -0,0 +1,369 @@
...
+    enabled: !!address,
+    select: (allowance) => {
+      return allowance < parseUnits(state.stakeNumber, GOV_TOKEN_DECIMALS);
+    },
+  });
+
+  // prepare approve
+  const { config: approveConfig } = usePrepareErc20Approve({
+    address: GOV_TOKEN_ADDRESS,
+    args: [STAKING_ADDRESS, parseUnits(state.stakeNumber, GOV_TOKEN_DECIMALS)],
```
t0rbik, line 110: if you input something that will break parseUnits whole component will throw run time error

components/stake/StakeFVM.tsx
```diff
@@ -0,0 +1,369 @@
...
+      refetchFvmBalance();
+      refetchStakedBalance();
+      refetchEarned();
+    },
+  });
+
+  /**
+   * Rewards related hooks
+   */
+  // we have only oFVM rewards
```
t0rbik, line 149: this is not 100% true bc maybe we will have something else. but for now only oFVM yes

components/stake/StakeFVM.tsx
```diff
@@ -0,0 +1,369 @@
...
+  useStakeFvmWithdraw,
+} from "../../lib/wagmiGen";
+import { LoadingSVG } from "../common/LoadingSVG";
+import { PRO_OPTIONS } from "../../stores/constants/constants";
+
+const buttonClasses =
+  "inline-flex items-center bg-blue/70 text-white hover:bg-blue/50 transition-colors duration-200 px-3 py-1 disabled:cursor-not-allowed";
+
+const inputClasses =
+  "w-full px-2 py-2 text-white hover:ring-1 hover:ring-cyan focus:outline-none focus:ring-1 focus:ring-cyan border-cyan-900/50 border bg-transparent";
```
t0rbik, line 32: this seems like overcomplication rather than optimisation. what if I need to change only one of many button classes for example

tailwind.config.js
oxSaturn: Definitely now finalized.

components/stake/StakeFVM.tsx
oxSaturn: Interesting, I can't reproduce it, both locally and vecel preview.

components/stake/StakeFVM.tsx
oxSaturn: IMO it's usually not a good idea to change styles of only one button when they're all on the same page. If you check the page, you can find I'm reusing (or repeating) styles as much as I can to keep consistency.

Would love to hear a specific example.

t0rbik: I won't give you exact examples, but instead why i believe this is an overkill.
Screenshot 2023-07-11 at 9 14 40 pm
Screenshot 2023-07-11 at 9 16 03 pm

oxSaturn: Usually I would prefer looping, but in this page it's not feasible.

Regarding the multi-cursor editing, I would say it's error-prone, let me give a contrived example, say I want to edit classes `relative w-full border border-[rgb(46,45,45)]` for two lists in a page, but it's reused somewhere between that two lists. Now I'll have to be very careful to not select it.

While in my case, hopefully I wouldn't reuse it at all as it's semantic, lol.

Anyway, I don't see this a big problem that we have to come with a conclusion. We can of course revisit those when issue rises.

components/stake/StakeFVM.tsx
oxSaturn: I've rewritten the code in https://github.com/Velocimeter/frontend/pull/257/commits/19038487e34d3680a38606b98e025f5e49b8398d to avoid hard coding rewards.

components/stake/StakeFVM.tsx
```diff
@@ -0,0 +1,423 @@
...
+      <div className="space-y-5">
+        <Section>
+          <SubHeader text="Stake" />
+          <div className="flex justify-between">
+            <span>Total Staked</span>
+            <span
+              className="underline cursor-pointer"
+              onClick={() => {
+                dispatch({ type: "stake", payload: fvmBalance?.formatted! });
+              }}
```
oxSaturn, line 217:
```suggestion
```

In my original design I used underline for clickable number so user can fills the input quickly, but this is not a clickable number, so we can just remove those code.

components/stake/StakeFVM.tsx
```diff
@@ -0,0 +1,423 @@
...
+            <span>{GOV_TOKEN_SYMBOL} balance</span>
+            <span
+              className="underline cursor-pointer"
+              onClick={() => {
+                dispatch({ type: "stake", payload: fvmBalance?.formatted! });
+              }}
+            >
+              {fvmBalance?.formatted}
+            </span>
+          </div>
```
oxSaturn, line 232: Do you think it would be better to move this balance below APR so the clickable number are more nearer to the input field.

CleanShot-PWCBenHY@2x

components/stake/lib/useApr.ts
```diff
@@ -0,0 +1,113 @@
...
+        tvl === undefined ||
+        !left ||
+        !discount ||
+        !veDiscount ||
+        rewardTokens.length !== rewardRates.length ||
+        rewardTokens.length !== left.length ||
+        !tokenPrices
+      ) {
+        throw new Error("Missing data");
+      }
```
oxSaturn, line 83: Really an interesting method employed here to make sure all data are ready.

oxSaturn (commented):
Awesome work! left some comments.

PR #258, opened 2023-07-11
summarise gauge option boost and option  emission aprs

t0rbik:
**Before:**
Screenshot 2023-07-11 at 9 30 49 pm

**After:**
Screenshot 2023-07-11 at 9 30 25 pm

---

### Detailed summary

- The code is modifying the `pairAprs` variable based on certain conditions.
- If the `filtered` array is not empty, it assigns the `filtered` array to `pairAprs`.
- If the `filtered` array is empty, it performs additional operations on the `filtered` array.
- It creates a `Map` object called `map`.
- It iterates over the `filtered` array and checks if each element has a property called `min_apr`.
- If it does, it checks if the `map` already has an entry for the symbol of the current element.
- If it does, it updates the `min_apr` and `max_apr` values of the existing entry in the `map`.
- If it doesn't, it adds a new entry to the `map` with the symbol and other properties of the current element.
- If an element in the `filtered` array does not have a `min_apr` property, it adds it directly to the `map`.
- It converts the values of the `map` to an array and assigns it to `pairAprs`.
- If the resulting array is empty, it assigns `null` to `pairAprs`.
- Finally, it assigns the `pairAprs` array to the `aprs` property of the `pair` object.

### DIFF
--- a/pages/api/pairs.ts
+++ b/pages/api/pairs.ts
@@ -36,7 +36,31 @@ export default async function handler(
             }
           }
         });
-        const pairAprs = filtered.length > 0 ? filtered : null;
+
+        // When gauge emits options and has been boosted with more options we want to summarise min and max apr values
+        const map = new Map<string, Exclude<Aprs, null>[number]>();
+        filtered.forEach((apr) => {
+          if ("min_apr" in apr) {
+            if (map.has(apr.symbol)) {
+              const current = map.get(apr.symbol) as Extract<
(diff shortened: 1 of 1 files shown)


oxSaturn (approved):
Looks good, would be great if we can have a unit test for the calculation.

oxSaturn:
@t0rbik Shall we apply a similar change to options page?

CleanShot-SeWxp3xM@2x

As there're two oFVM APR.

Issue #193, opened 2023-07-02
Vote value flick on "Cast Vote" click


t0rbik:
**Current Behaviour**
When i select a pair I want to vote for on votes page, after I click `Cast Vote`. It flicks my selected vote back to previous number and causes me to warn: am I casting the vote as I selected?

**Expected Behaviour**
After I select a vote and click on `Cast Vote` selected vote percent doesn't flick back to previous value.

**Steps To Reproduce**
1. Go to vote page.
2. Select a vote amount for any pair.
3. Click on `Cast Vote`.
4. You will see your selected value flick back to initial position.

t0rbik:
If you have 1+ nft then nft flicks after the vote. maybe we can take a look into it as well?

oxSaturn:
> If you have 1+ nft then nft flicks after the vote. maybe we can take a look into it as well?

Say we have two nfts, A and B, both votable in this epoch.

And we vote with A, the dropdown will show B automatically once A vote is done. Is this the same thing you describe?

t0rbik:
Yes

oxSaturn:
> Yes

That would be a feature, not a bug lol. I think my plan was to save users a click, otherwise you'll have to select another nft to vote while currently it will select the next nft for you automatically.

Want to disable it? I agree it could be confusing.

t0rbik:
I actually like it, but it can cause confusion if you cast vote. then your vote will flick to show `next nft` votes. Besides, one may think that he didnt cast, recast and now he voted with both of his nfts. I think it's good but in the end can cause confusion. It's not a high priority tho I imagine. I reckon not many users have multiple nfts and if they do so tend to megre into 1.

oxSaturn:
I believe I encountered another flickering bug when voting.

1. casted my new votes
2. confirmed in my wallet
3. the transaction was pending
4. saw the votes reverting to old ones first
5. after transaction was finalized, votes reverted again to the new ones.

I'm wondering if the `pairs` in `queryKey` might cause the problem:

![CleanShot-PkkYHShW](https://github.com/Velocimeter/frontend/assets/126733611/b5f99873-1650-43a2-a21d-3e934708fbdf)

I.e., the `pairs` data got updated during my voting, since the `queryKey` was changed, `useQuery` would issue a new request, causing the flickering problem.

PR #270, opened 2023-07-14
Fix votes' flickering

oxSaturn:
Closes https://github.com/Velocimeter/frontend/issues/193

---

### Focus of the PR:
This PR focuses on adding the `refetchOnWindowFocus` option to the `queries.ts` file in the `ssVotes` component.

### Detailed summary:
- Added the `refetchOnWindowFocus` option with a value of `false` to the `enabled` object in the `queries.ts` file.

### DIFF
--- a/components/ssVotes/lib/queries.ts
+++ b/components/ssVotes/lib/queries.ts
@@ -28,6 +28,7 @@ export const useVestVotes = (tokenID: string | undefined) => {
     queryFn: () => getVestVotes(address, tokenID, pairs),
     enabled: !!address && !!tokenID && !!pairs,
     refetchOnMount: false,
+    refetchOnWindowFocus: false,
     select: (votes) => {
       const votesValues: Votes | undefined = votes?.map((vote) => {
         return {


components/ssVotes/lib/queries.ts
```diff
@@ -28,6 +28,7 @@ export const useVestVotes = (tokenID: string | undefined) => {
     queryFn: () => getVestVotes(address, tokenID, pairs),
     enabled: !!address && !!tokenID && !!pairs,
     refetchOnMount: false,
+    refetchOnWindowFocus: false,
```
oxSaturn, line 31: Since user might change the votes and switch the focus to other windows before casting, it doesn't make sense to refetch the votes data on window focus, as it would discard user's new votes.

oxSaturn:
Ok, I reckon it's really that simple to fix.

The problem here is when we cast votes, we open the confirmation dialog from wallet, which means the current page will lose focus, since the `refetchOnWindowFocus` defaults to `true`, the page will refetch the old votes once gaining back focus. And our voting transaction might still be pending, refetch would only get the old data instead of new ones.

t0rbik:
can you set up same to canto? i have dust nft there to look

PR #271, opened 2023-07-14
Add completely killed gauges 

t0rbik:
add killed gauges data from backend and possibility to withdraw from …
API url to test https://p01--test-api--wlcfywkylwkq.code.run

---

### Detailed summary
- Added `@radix-ui/react-select` package to `package.json`
- Added `oBLOTR` object with `tokenAddress` property to `wagmi.config.ts`
- Updated import paths in `pages/liquidity/[address]/index.tsx`
- Updated `pair` variable assignment in `components/liquidityManage/lib/mutations.ts`
- Updated import paths in `components/options/lib/useAllowance.ts`
- Updated `onView` function in `components/liquidityPairs/LiquidityPairsTable.tsx`
- Updated import paths in `components/options/lib/useIsEmittingOptions.ts`
- Updated import paths in `components/stake/lib/useApr.ts`
- Updated import paths in `components/options/stake.tsx`
- Updated import paths in `components/liquidityManage/LiquidityManage.tsx`
- Updated import paths in `components/options/reward.tsx`
- Updated import paths in `components/options/lib/useDiscountsData.ts`
- Updated import paths in `components/options/lib/useGaugeRewards.ts`
- Updated import paths in `components/common/radixTooltip.tsx`
- Updated import paths in `components/common/radixSelect.tsx`

### Focus of the PR
This PR focuses on updating import paths and adding new dependencies to the project.

> The following files were skipped due to too many changes: `components/common/radixSelect.tsx`, `stores/types/types.ts`, `lib/global/queries.ts`, `components/options/lib/useStakeData.ts`, `components/options/lib/useTokenData.ts`, `components/liquidityManage/lib/queries.ts`, `components/options/lib/useAmountToPay.ts`, `components/options/lib/useGaugeApr.ts`, `pnpm-lock.yaml`, `components/options/redeem.tsx`, `stores/abis/voterABI.ts`


oxSaturn:
CleanShot-CuLnC7aQ@2x

Might want to fix the style here.

---

I was not in the discussion with your guys so I could be wrong here, but from the code I can see that a pair could have no active gauge, or one active gauge, and multiple killed gauge.

So here's my suggestion:

Currently we're showing 2 separate rows for two gauges that belongs to one pair, and I reckon we can take advantage of row span to improve how we show those killed gauges here:

CleanShot-p9rIkhN3@2x

We'll be able to show as many gauges as we can and still keep the page very intuitive.

Besides, how about introducing a nest route to handle different gauge here? E.g., `/liquidity/[address]/[...gauge]`, having button like blow is sort of confusing IMO:

CleanShot-qzE9qX9I@2x

t0rbik:
for the rows span. it is currently set to hide killed gauges at all, and only show if you have balance to withdraw. plus we dont want to spook people who already in "right" gauge and make them check what happened and why do they see a sign. and also we want for those who are in "outdated" gauge has an easy-way-no-panic way to see their assets. i find separate row which will just go away after withdraw the easiest way. but we can dig and see other ways as well as proposed. 
nested route will do. styles yeh

oxSaturn:
> for the rows span. it is currently set to hide killed gauges at all, and only show if you have balance to withdraw. plus we dont want to spook people who already in "right" gauge and make them check what happened and why do they see a sign. and also we want for those who are in "outdated" gauge has an easy-way-no-panic way to see their assets. i find separate row which will just go away after withdraw the easiest way. but we can dig and see other ways as well as proposed. nested route will do. styles yeh

With row spans, we can still hide them for users who have no balance in those killed gauges. Definitely don't want to show them always.

t0rbik:
Could you look at it with separate rows for now?

Maybe you could do a PR into this PR or into fantom to utilise nested rows? We would want to show killed only if user has balance in it. And warning icon only if has balance in it.

oxSaturn:
> Could you look at it with separate rows for now?
> 
> Maybe you could do a PR into this PR or into fantom to utilise nested rows? We would want to show killed only if user has balance in it. And warning icon only if has balance in it.

Let me try. I'll file another PR based on this one, believe I have no permission to push to this branch.

t0rbik:
Merging this because scanto/blotr gauge has been killed. Need to show up killed gauge data asap.

PR #278, opened 2023-07-16
Feat/killed gauges

oxSaturn:
Based on https://github.com/Velocimeter/frontend/pull/271


lib/global/queries.ts
```diff
@@ -588,7 +588,15 @@ export const getPairsWithGauges = async (
     return pair;
   });
 
-  return [...ps1, ...ps2];
+  for (const pair of ps2) {
+    const findPair = ps1.find((p) => p.address === pair.address);
+    if (findPair) {
+      // @ts-ignore FIXME here, we change the type of killed_gauges
+      findPair.killed_gauges = [pair];
```
oxSaturn, line 595: Here I've attached the data in `ps2` to respective pair in `ps1` so we can show them later.

components/liquidityPairs/LiquidityPairsTable.tsx
```diff
@@ -843,99 +813,153 @@ function Row(props: { row: Pair; onView: (_row: Pair) => void }) {
...
+                className="text-xs font-extralight"
+                noWrap
+                color="textSecondary"
+              >
+                {row?.stable ? "Stable Pool" : "Volatile Pool"}
+              </Typography>
+            </div>
+          </div>
+        </TableCell>
+        <GaugeRow row={row} onView={onView} />
```
oxSaturn, line 940: A new `GaugeRow` component is introduced to show all columns other than the first two.

lib/global/queries.ts
t0rbik: can we add another field?

t0rbik: instead of mutating?

oxSaturn: Sure, let me fix a bug first, lol.

oxSaturn:
Ok, just noticed this pr has a bug https://github.com/Velocimeter/frontend/pull/271 which can be reproduced in this PR.

Say I'm visiting page like `/liquidity/address/killed_gauge` directly, this `pairs` will become `undefined`:

https://github.com/Velocimeter/frontend/blob/85514b0bb2e8048f87f1647351a8fbefbe377eb5/components/liquidityManage/lib/queries.ts#L233-L241

So the code after will continue to run, and we'll get a balance from the alive gauge. After a while, the `pairs` gets filled with data, and the balance will get updated as well.


https://github.com/Velocimeter/frontend/assets/126733611/3dfd7b1e-2100-4ddd-893c-c8ad217c61da

Check the video above, the balance will change from 0.13 to 7.98 after a while.

oxSaturn:
The problem of this PR is I can no longer find `pair` data for my killed gauge as they don't exist in `pairs`, and since the code to get `thePair` is broken for killed gauge, thus it won't work.

t0rbik:
So how about we wait for pairs to load before we run get pair by address first time?

oxSaturn:
> So how about we wait for pairs to load before we run get pair by address first time?

lol, it will work for the old case, but not case in this pr, my code would still need to rely on the fallback code to get the data.

components/liquidityPairs/LiquidityPairsTable.tsx
```diff
@@ -843,99 +813,152 @@ function Row(props: { row: Pair; onView: (_row: Pair) => void }) {
...
+                {row?.stable ? "Stable Pool" : "Volatile Pool"}
+              </Typography>
+            </div>
+          </div>
         </TableCell>
+        <GaugeRow row={row} onView={onView} />
       </TableRow>
+      {/* show killed gauges only if user has any balance in it */}
+      {(row.dead_pairGauges ?? [])
+        .filter((pair) => {
```
t0rbik, line 944:
```suggestion
      {row.dead_pairGauges
        ?.filter((pair) => {
```

t0rbik: I think this more makes sense

components/liquidityPairs/LiquidityPairsTable.tsx
t0rbik, line 1: i think it looks very good. Just very much wondering and trying to find a way to not to delete background on hover. and moreover to add it to whole row span when hovered

stores/types/types.ts
```diff
@@ -147,6 +147,7 @@ interface Pair {
     | "bribeAddress"
     | "bribes"
   >[]; // if no gauges killed, set to: []
+  dead_pairGauges?: Pair[]; // this is a dirty hack which I don't like, we might need to refactor the types here
```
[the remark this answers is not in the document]
t0rbik: or even better just `workaround`

t0rbik:
Do you think there are changes to make to descending comparator as well?

components/liquidityPairs/LiquidityPairsTable.tsx
t0rbik: But actually we can leave this as it for now

lib/global/queries.ts
```diff
@@ -588,7 +588,14 @@ export const getPairsWithGauges = async (
...
 
-  return [...ps1, ...ps2];
+  for (const pair of ps2) {
+    const findPair = ps1.find((p) => p.address === pair.address);
+    if (findPair) {
+      findPair.dead_pairGauges = [{ ...pair }];
+    }
+  }
+
+  return ps1;
```
t0rbik, line 598: in this variant need to think how to still read and update accordignly getRewardBalances

oxSaturn: So our `Rewards` page isn't showing rewards of those killed gauges at the moment?

oxSaturn:
Reckon we'd better keeping types as simple as possible and only introduce new field when we must have to, otherwise we can easily get a huge type that's hard to maintain. Many data could just be derived from other fields.

Take the `Pair` type for example, we may maintain an array of gauge:

```

interface GaugeX {
  state: 'active' | 'killed'
  address: `0x${string}`;
  // ...
}
interface Pair {
  gauges: GaugeX[];
  tvl: number;
  address: `0x${string}`;
  // ...
}
```
And data like `gauge_address` or `isAliveGauge` can be derived from `gauges`, and hopefully we can eliminate fields like `killed_gauges` or `dead_pairGauges` as well.

lib/global/queries.ts
t0rbik: It is showing in prod because we add separate pair to pairs with gauges so it will check its gauge. but if we add it to filed on pair it wont check it right now

t0rbik:
This would require a lot of code changes not only in frontend but in backend. I agree that Pair type can be improved. However, when choosing between overwriting existing fields _in place_ i'd rather add new filed - this regarding the dead_killed_gauges field we mentioden here.

oxSaturn:
> So how about we wait for pairs to load before we run get pair by address first time?

Was trying to update code for rewards, but found there might be something wrong here as the `pairs` kept returning `undefined` when testing locally:

CleanShot-eGtkwYlf

CleanShot-w2dZE7JD@2x

I.e., I was stuck on this loading screen:

CleanShot-1NqdMLOL@2x

oxSaturn:
> > So how about we wait for pairs to load before we run get pair by address first time?
> 
> Was trying to update code for rewards, but found there might be something wrong here as the `pairs` kept returning `undefined` when testing locally:
> 
> CleanShot-eGtkwYlf
> 
> CleanShot-w2dZE7JD@2x
> 
> I.e., I was stuck on this loading screen:
> 
> CleanShot-1NqdMLOL@2x

Very interesting, seems the code of mutating `dead_pairGauges` is causing the problem. Could it be react query doesn't like mutations? I don't know.

stores/abis/gaugeABI.ts
```diff
@@ -1,41 +1,13 @@
 export const gaugeABI = [
   {
     inputs: [
-      {
```
oxSaturn, line 4: I've updated the abi per https://ftmscan.com/address/0x38ed4dc09a7a810053eac923bff4d1d8c2cf4d62#code, i.e., Gauge V2.

components/liquidityManage/lib/queries.ts
```diff
@@ -433,9 +433,8 @@ export const getPairByAddress = async (
             args: [account],
           },
           {
-            ...gaugesContract,
-            functionName: "external_bribes",
-            args: [gaugeAddress],
+            ...gaugeContract,
+            functionName: "external_bribe",
```
oxSaturn, line 437: I believe reading `external_bribes` from voter contract is wrong here, as it will return `0x0000000000000000000000000000000000000000` for killed gauge.

oxSaturn:
> Do you think there are changes to make to descending comparator as well?

Nah, I don't think it's a problem at all.

components/liquidityManage/lib/queries.ts
t0rbik: we cant read external bribe from gauge as mentioned here https://github.com/Velocimeter/frontend/pull/281#issuecomment-1637080108

stores/abis/gaugeABI.ts
t0rbik: we already have gauge v3, it is backwords compatible. also remember that we are not supposed to read external_bribe from gauge.

components/ssRewards/lib/queries.ts
```diff
@@ -52,6 +52,13 @@ export const getRewardBalances = async (
     throw new Error("pairs not found");
   }
 
+  pairs = pairs.reduce((acc, pair) => {
+    if (pair.dead_pairGauges && pair.dead_pairGauges.length > 0) {
+      return acc.concat(pair).concat(pair.dead_pairGauges);
+    }
+    return acc.concat(pair);
+  }, [] as Pair[]);
```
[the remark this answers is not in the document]
oxSaturn: because I got bitten by https://github.com/Velocimeter/frontend/pull/278#issuecomment-1639456385, thus prefer to use immutable data structure here. lol

components/liquidityManage/lib/queries.ts
t0rbik: what is returned by gauge address?

t0rbik: why do you need external bribe for killed gauge?

t0rbik: The external bribe of killed gauge is the same as external bribe of new gauge because we setExternalBribe back in voter.
https://discord.com/channels/1060312715496783952/1060314450659377222/1130036136518090783

components/ssRewards/lib/queries.ts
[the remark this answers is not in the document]
oxSaturn: Didn't try it yet. Do u have performance concern here? Usually I don't pay attention to it unless I see performance degrade.

components/liquidityManage/lib/queries.ts
oxSaturn: Here's the problem I have when visiting my LP with a killed gauge:

CleanShot-2SPGotW1@2x

CleanShot-begE0837@2x

Can you fetch this PR and revert changes here to see if you can reproduce?

t0rbik: will do

components/ssRewards/lib/queries.ts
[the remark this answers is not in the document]
t0rbik: It just logically hits my mind that we are going to allocate memory on each iteration to new array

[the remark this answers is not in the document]
oxSaturn: lol, no worries sir, I believe those can be garbage collected easily. Have u heard https://immutable-js.com/, or Elixir where data are immutable by default.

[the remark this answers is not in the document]
t0rbik: now, i found the piece of information that is telling this thoughts inside my head:
[piece of information](https://youtube.com/shorts/Q-jnCu85PEc?feature=share)

components/liquidityManage/lib/queries.ts
t0rbik, line 1: 1. I recall wth your implementation of data the killed gauge will always pass through to fetching from chain (you probably already know that)

t0rbik: 2. We in fact don't need any bribe data related to killed gauge and that's good that voter is sending zero address, so that we can't really call some other bribe contract different from connected to new gauge

t0rbik: 3. I guess if you want to leave it as pure Pair type you would fetch new gauge address from voter using pair address, fetch bribe address for new gauge address and assign to killed gauge address

components/ssRewards/lib/queries.ts
t0rbik, line 1: this doesnt seem to make any harm
```const pairsToCheckRewards: Pair[] = [];

  pairs.forEach((pair) => {
    if (pair.dead_pairGauges && pair.dead_pairGauges.length > 0) {
      pairsToCheckRewards.push(pair);
      pairsToCheckRewards.push(...pair.dead_pairGauges);
    } else {
      pairsToCheckRewards.push(pair);
    }
  });

  const gauges = pairsToCheckRewards.filter(hasGauge);
```

stores/abis/gaugeABI.ts
```diff
@@ -256,13 +256,6 @@ export const gaugeABI = [
     stateMutability: "view",
     type: "function",
   },
-  {
-    inputs: [],
-    name: "external_bribe",
-    outputs: [{ internalType: "address", name: "", type: "address" }],
-    stateMutability: "view",
-    type: "function",
-  },
```
oxSaturn: Removing this will cause many type errors.

CleanShot-sulBMlQW@2x

components/liquidityManage/lib/queries.ts
oxSaturn: > 1. I recall wth your implementation of data the killed gauge will always pass through to fetching from chain (you probably already know that)

Right. I already knew that. Actually that's how I ran into the issue.

oxSaturn: > 2. We in fact don't need any bribe data related to killed gauge and that's good that voter is sending zero address, so that we can't really call some other bribe contract different from connected to new gauge

I don't fully understand this. Users still might have rewards in killed gauges, right? Doesn't that mean we still need bribe data to calculate their rewards?

components/liquidityManage/lib/queries.ts
```diff
@@ -229,15 +229,22 @@ export const getPairByAddress = async (
...
-    }
-    return p.address.toLowerCase() === pairAddress.toLowerCase();
-  });
+  const pair = pairs
+    ?.reduce((acc, pair) => {
+      if (pair.dead_pairGauges) {
+        return [...acc, pair, ...pair.dead_pairGauges];
+      }
+      return [...acc, pair];
+    }, [] as Pair[])
```
oxSaturn, line 238: Ok, this is the same trick I employed for Rewards, since it's transient data, so it won't have problematic side effects.

components/liquidityManage/lib/queries.ts
```diff
@@ -433,8 +440,9 @@ export const getPairByAddress = async (
             args: [account],
           },
           {
-            ...gaugeContract,
-            functionName: "external_bribe",
+            ...gaugesContract,
+            functionName: "external_bribes",
+            args: [gaugeAddress],
```
oxSaturn: Reverted.

components/ssRewards/lib/queries.ts
oxSaturn: lol, I thought the video you shared was telling us something like "guys, you're writing UI with React which sometimes re-renders like insanely, why would you care about the performance of immutable data?`

Anyway, if you still prefer `.push`, I can update them. Usually I don't give attention to the syntax we use as most of the time, the performance bottle won't be them.

components/liquidityManage/lib/queries.ts
oxSaturn: Ok, I sort of understand the situation now. The killed gauges once had their own bribe contracts, but they were nullified when killed, and the bribe contract was inherited by the live gauge.

PR #281, opened 2023-07-16
DO NOT USE EXTERNAL BRIBE FROM GAUGE CONTRACT

t0rbik:
### Detailed summary

- Updated the `createGauge` mutation to pass `1n` instead of `0n` as the second argument.
- Removed the `external_bribe` function from `gaugeABI.ts` and `maxxingGaugeABI.ts`.
- Added `gaugeAddress` property to the `Vote` type in `types.ts`.
- Updated the `getVestVotes` function in `queries.ts` to include the `gaugeAddress` property in the returned object.
- Updated the `getPairsWithGaugesAndVotes` function in `queries.ts` to filter based on both `address` and `gaugeAddress`.
- Removed the `bribeContractInstance` and related code from `queries.ts`.
- Updated the `getPairByAddress` function in `queries.ts` to use `viemClient` for reading contract data.
- Updated the `ssVotesTable.tsx` component to include `row.gauge.address` in the `key` prop.
- Updated the `VotesRow` component in `ssVotesTable.tsx` to retrieve the `sliderValue` based on `row.address`.
- Updated the `descendingComparator` function in `ssVotesTable.tsx` to retrieve the `sliderValue` based on `a.address` and `b.address`.
- Updated the `descendingComparator` function in `ssVotesTable.tsx` to retrieve the `sliderValue` based on `a.address` and `b.address`.

### DIFF
--- a/components/liquidityManage/lib/mutations.ts
+++ b/components/liquidityManage/lib/mutations.ts
@@ -813,7 +813,7 @@ const writeCreateGauge = async (
       address: CONTRACTS.VOTER_ADDRESS,
       abi: CONTRACTS.VOTER_ABI,
       functionName: "createGauge",
-      args: [pairAddress, 0n],
+      args: [pairAddress, 1n],
     });
     const txHash = await walletClient.writeContract(request);
     return txHash;
--- a/components/liquidityManage/lib/queries.ts
+++ b/components/liquidityManage/lib/queries.ts
@@ -3,7 +3,6 @@ import { Address, useAccount } from "wagmi";
 import {
   formatEther,
   formatUnits,
-  getContract,
   isAddress,
   parseEther,
   parseUnits,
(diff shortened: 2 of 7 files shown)


components/liquidityManage/lib/mutations.ts
```diff
@@ -813,7 +813,7 @@ const writeCreateGauge = async (
       address: CONTRACTS.VOTER_ADDRESS,
       abi: CONTRACTS.VOTER_ABI,
       functionName: "createGauge",
-      args: [pairAddress, 0n],
+      args: [pairAddress, 1n],
```
oxSaturn, line 816: Is there any limit for this second argument? I saw the contract accepts `unit256`, so why choose `1n` over `0n` or why not use `3n`?

t0rbik: because we deployed new gauge factory and added new gauge factory to voter. from now one we need all new gauges be of new type. new gauge type supports velocimeter pro

oxSaturn: Which means we'll use `3n` if we deployed another new gauge factory? And so forth.

oxSaturn (approved):
Code looks good, reckon I'll need to learn a little smart contract so I can know wtf is this external bribe.

components/liquidityManage/lib/mutations.ts
t0rbik: yes. you can look at the voter in v3-contracts

t0rbik:
external bribe is the contract which receives bribes. it also holds the votes data inside itself. we have a mapping of external bribes on voter. we can set to whatever we want there. in our case we are switching gauge of existing pair with existing bribe contract. so we need to bring that contract back. we can do it in voter. but gauge also has this data which is immutable. It is not being used in fact in contracts codebase and we must do not use it anywhere else bc as it stands it can give inaccurate data. and that's dangerous to mess with this contract address.

PR #284, opened 2023-07-17
do not auto select next votable nft

oxSaturn:
Related issue https://github.com/Velocimeter/frontend/issues/193#issuecomment-1637365612

---

### Detailed summary
- Added a condition to prevent changing the selected token if it has already been selected by the user.
- Set the same token to update its data if it has been selected.
- Updated the dependency array of the second useEffect hook to include the `token` variable.


t0rbik (approved):

t0rbik:
can we have same to fantom and pulse

PR #288, opened 2023-07-18
add underline to clickable numbers

oxSaturn:
Per discussion with dunks in the latest discord issue.

---

## PR-Codex overview
This PR focuses on updating the stake functionality in the options component. 

### Detailed summary
- Updated the condition for setting the stake amount based on staked balance without lock.
- Added underline style to the staked without lock amount when action is "withdraw".
- Removed cursor-pointer and onClick event from staked with lock amount.
- Added underline style to pooled balance when action is "stake" and added onClick event to set the maximum amount.

### DIFF
--- a/components/options/stake.tsx
+++ b/components/options/stake.tsx
@@ -161,8 +161,11 @@ export function Stake() {
         setAmount(pooledBalance.formatted);
       }
     } else {
-      if (stakedBalance && parseFloat(stakedBalance) > 0) {
-        setAmount(stakedBalance);
+      if (
+        stakedBalanceWithoutLock &&
+        parseFloat(stakedBalanceWithoutLock) > 0
+      ) {
+        setAmount(stakedBalanceWithoutLock);
       }
     }
   };
@@ -268,7 +271,9 @@ export function Stake() {
         <div className="flex items-center justify-between">
           <div>Staked without lock</div>
           <div
-            className={`${action === ACTION.WITHDRAW && "cursor-pointer"}`}
+            className={`${
+              action === ACTION.WITHDRAW && "cursor-pointer underline"
+            }`}
             onClick={() => pickWithdrawAmount("notLocked")}
           >
             ${formatCurrency(stakedWithoutLockValue)}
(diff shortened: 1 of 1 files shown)


components/options/stake.tsx
```diff
@@ -161,8 +161,11 @@ export function Stake() {
         setAmount(pooledBalance.formatted);
       }
     } else {
-      if (stakedBalance && parseFloat(stakedBalance) > 0) {
-        setAmount(stakedBalance);
+      if (
+        stakedBalanceWithoutLock &&
+        parseFloat(stakedBalanceWithoutLock) > 0
+      ) {
+        setAmount(stakedBalanceWithoutLock);
```
oxSaturn, line 168: Since users can't withdraw the staked balance with lock, I think it might be better just allowing populating of staked balance without lock.

components/options/stake.tsx
```diff
@@ -268,20 +271,17 @@ export function Stake() {
...
         </div>
         <div className="flex items-center justify-between">
           <div>Staked with lock</div>
-          <div
-            className={`${action === ACTION.WITHDRAW && "cursor-pointer"}`}
-            onClick={() => pickWithdrawAmount("locked")}
-          >
-            ${formatCurrency(stakedWithLockValue)}
-          </div>
+          <div>${formatCurrency(stakedWithLockValue)}</div>
```
oxSaturn, line 284: Make locked value not clickable, i.e., can't be populated.

components/options/stake.tsx
```diff
@@ -291,7 +291,14 @@ export function Stake() {
...
         <div className="flex items-center justify-between">
           <div>Pooled balance</div>
-          <div>{pooledBalance?.formatted}</div>
+          <div
+            className={`${
+              action === ACTION.STAKE && "cursor-pointer underline"
+            }`}
+            onClick={() => setMax()}
+          >
+            {pooledBalance?.formatted}
```
oxSaturn, line 300: Make pooled balance clickable. I think the Max button works great sometimes, but it can be confusing when there are multiple numbers that can be populated on the page. Underlined text makes it very clear that you're populating that specific number into the input field.

t0rbik (approved):

PR #301, opened 2023-07-20
do not refetch votes when voting

oxSaturn:
Related issue https://github.com/Velocimeter/frontend/issues/193#issuecomment-1643868889

When users are voting, we should avoid refetching the votes data, otherwise the votes would revert to the old ones first causing confusion.

---

## PR-Codex overview
This PR focuses on adding a `useIsVoting` hook to manage the voting state in the `ssVotes` component. 

### Detailed summary
- Added `useIsVoting` hook in `components/ssVotes/lib/useIsVoting.ts` to manage the `isVoting` state.
- Imported and used `useIsVoting` hook in `components/ssVotes/lib/mutations.ts` and `components/ssVotes/lib/queries.ts` to set and check the voting state.


components/ssVotes/ssVotes.tsx
```diff
@@ -96,7 +97,10 @@ export default function Votes() {
     setShowWarning(false);
   }, []);
 
+  const setIsVoting = useIsVoting((state) => state.setIsVoting);
+
   const onVote = () => {
+    setIsVoting(true);
```
t0rbik, line 103: i think you can add it to onMutate callback in useVote, or as a callback to mutation fn as well (not sure about latter). Although this works too

oxSaturn: Reckon the current implement might makes more sense, i.e., the votes should stay intact after user clicked the `Cast Votes`. As it could still get updated between when user clicking the button and when `onMutate` gets fired?

t0rbik: hm how is it going to be updated if onMutate fires moment you click button, because clicking button fires mutation

t0rbik: onMutate kind of makes more sense because we already use onError and onSuccess

oxSaturn: > hm how is it going to be updated if onMutate fires moment you click button, because clicking button fires mutation

Ok, I was wrong, thought `onMutate` would only be fired once user confirms the transaction. But seems it will be fired right after clicking.

t0rbik (approved):
looks good

